PR #3, opened 2018-09-21
Initial check-in with syntax coloring.

allevato:
This took longer than I hoped, since I kept getting pulled away to work on other things. The main goal here was to do more than just copy the MagicPython grammar, but to strip it down to just what we needed for Starlark (and to add a couple additional enhancements to bring it up to a similar level of highlighting that VS Code uses for other languages).

### DIFF
--- a/syntaxes/starlark.tmLanguage.json
+++ b/syntaxes/starlark.tmLanguage.json
@@ -1,0 +1,929 @@
+{
+    "name": "Starlark",
+    "scopeName": "source.starlark",
+    "fileTypes": [
+        "bzl"
+    ],
+    "patterns": [
+        {
+            "include": "#statement"
+        },
+        {
+            "include": "#expression"
+        }
+    ],
+    "repository": {
+        "statement": {
+            "patterns": [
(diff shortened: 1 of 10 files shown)


syntaxes/starlark.tmLanguage.json
```diff
@@ -0,0 +1,929 @@
...
+                        }
+                    }
+                }
+            ]
+        },
+        "builtin-function": {
+            "patterns": [
+                {
+                    "name": "support.function.starlark",
+                    "match": "(?<!\\.)\\b(all|any|bool|dict|dir|enumerate|getattr|hasattr|hash|int|len|list|load|max|min|print|range|repr|reversed|sorted|str|tuple|type|zip)\\b"
```
laurentlb: `load` can be removed, since it matches `builtin-constant`?

allevato: `load` is here so that those statements have their arguments colored consistently with other functions (which makes a difference when you use aliasing via keyword arguments).

PR #13, opened 2018-11-27
Update package-lock.json to remove security vulnerability.

allevato:
This backs us off event-stream 3.3.6, which added a dependency on
flatmap-stream which contained malicious code (see
https://github.com/dominictarr/event-stream/issues/116). Our
dependency was locked to a version of flatmap-stream that does not
seem to have been affected, but this removes the dependency entirely
to be safe.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,7 +70,7 @@
         },
         "@types/events": {
             "version": "1.2.0",
-            "resolved": "http://registry.npmjs.org/@types/events/-/events-1.2.0.tgz",
+            "resolved": "https://registry.npmjs.org/@types/events/-/events-1.2.0.tgz",
             "integrity": "sha512-KEIlhXnIutzKwRbQkGWb/I4HFqBuUykAdHgDED6xqwXJfONCjF5VoE0cXEiurh3XauygxzeDzgtXUqvLkxFzzA==",
             "dev": true
         },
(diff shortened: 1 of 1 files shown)


thomasvl:
@allevato a bunch of the `resolved` urls went from _https_ to _http_ which seems a little odd.  Any idea why?

allevato:
It looks like a known issue in the registry itself (ironic timing, given what this commit is meant to fix): https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285

The suggested workaround for now is to just replace `s/http:/https:/` because they're still hosted there, so I'll do that.

laurentlb:
We still have a security issue:

> Vulnerable versions: > 3.3.4
> Patched version: No fix
> The NPM package flatmap-stream is considered malicious. A malicious actor added this package as a dependency to the NPM event-stream package in versions 3.3.6 and later. Users of event-stream are encouraged to downgrade to the last non-malicious version, 3.3.4.
> 
> Users of flatmap-stream are encouraged to remove the dependency entirely.
> 

https://github.com/bazelbuild/vscode-bazel/blob/master/package-lock.json#L469

allevato:
It's not entirely clear to me (as I'm not much of a Node expert) how to resolve alerts like these if the intermediate dependencies we're using are the ones pulling in a specific version. We don't depend directly on `event-stream`; it's a few layers deep in the graph.

laurentlb:
Do you know which package depends on it?
They probably have an issue about it, that we can follow.

PR #21, opened 2018-12-05
Enforce tslint and reformat code.

allevato:
This adds support for `tslint` to our code base and reformats all of our code to comply to its `tslint:recommended` rule set.

The most major changes here are **one class per source file** and **80 columns instead of 100**.

After this is in, I plan to set up Travis to run lint checks so that we can keep enforcement consistent and automatic for contributions.


laurentlb:
The commit "Reformat all files and fix lint warnings." seems to add a lot of code. Can you double-check it?

e.g. "src/bazel/bazel_child_process_command.ts", "src/workspace-tree/bazel_workspace_tree_provider.ts"

allevato:
Those additions are really just one large file being split into one-class-per-file to address that particular lint error (e.g., see workspace-tree.ts which was deleted, and the code split out into the other files in that directory). In retrospect I could have done that in a separate commit, but since I addressed all the lint errors at once, that change got lumped in with some of the others.

PR #34, opened 2018-12-10
Add recommended extensions to vscode-bazel workspace.

dierksen:


.vscode/extensions.json
```diff
@@ -0,0 +1,14 @@
+{
+	// See http://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
```
smukherj1: Does the extension still work? JSON doesn't natively support comments which is why the diff appears all red on github

allevato: Yes, some of VS Code's JSON parsing is more lenient than the actual JSON spec, and it's part of their built-in templates for these files. 🙁

smukherj1: ¯\_(ツ)_/¯

alright ignore what I said then :+1:

PR #36, opened 2018-12-10
Add test option for test_suite targets.

dierksen:


src/workspace-tree/bazel_target_tree_item.ts
```diff
@@ -66,7 +66,8 @@ export class BazelTargetTreeItem
   }
 
   public getContextValue(): string {
-    if (this.queriedRule.ruleClass.endsWith("_test")) {
+    const rc = this.queriedRule.ruleClass;
```
allevato: Please use unabbreviated variable names, like `ruleClass`.

PR #41, opened 2018-12-16
Migrate Bazel build/test invocations to tasks.

### LINKED ISSUE #7: Use tasks for build/test actions
For expediency to get the debugging up and running, I've implemented commands that just launch Bazel directly to perform build/test runs. We should look into using the tasks APIs for these.

### LINKED ISSUE #20: Running build target fails with fish shell
With the default shell set to [fish](https://fishshell.com) building targets from the `BAZEL BUILD TARGETS` tree does not work.  The `&&` characters that are used in the terminal command `cd <package directory> && bazel build <target>` are not valid in the fish shell. Fish uses the syntax `cd <package directory>; and bazel build <target>`.

allevato:
Fixes #7.
Fixes #20.


src/bazel/tasks.ts
```diff
@@ -0,0 +1,63 @@
...
+      break;
+    case "test":
+      commandDescription = "Test";
+      break;
+  }
+
+  const targetsDescription = options.targets.join(", ");
+  return new vscode.Task(
+    { type: "bazel", command, targets: options.targets },
+    // TODO(allevato): Change Workspace to Global once the fix for
```
thomasvl: I don't suppose the vscode version is expose in a way to test to the code can be ready in advance?

allevato: I'd have to check to see if there are recommended best practices for how to toggle features based on version, or if there is a way to just require a particular minimum version of VS Code for the extension (it may be tied to the `vscode` version used in package.json).

PR #44, opened 2018-12-17
Add an explicit refresh to the Bazel tree view.

### LINKED ISSUE #26: Bazel Target Tree view should have a refresh command
An explicit refresh button in the view's toolbar would be useful if the user makes a change in their workspace that VS Code doesn't detect as requiring a refresh. For example, we refresh on changes in `BUILD` files, but not for `.bzl` files. Even if we do decide to start refreshing based on more changes, the explicit control is a good escape hatch if the IDE doesn't catch up.

allevato:
This is useful when manipulating files in the workspace in a way that
the file watcher doesn't detect, but which cause the build targets to
change.

Fixes #26.


package.json
```diff
@@ -47,6 +47,14 @@
                 "category": "Bazel",
                 "command": "bazel.clean",
                 "title": "Clean"
+            },
+            {
+                "command": "bazel.refreshBazelBuildTargets",
+                "title": "Refresh Bazel Build Targets",
```
thomasvl: Should this use "category": "Bazel" and the title change accordingly like all the others?

allevato: Yep, that was a merge resolution error. Thanks!

PR #45, opened 2018-12-17
Various debugger fixes.

### LINKED ISSUE #24: Breakpoints may or may not fire depending on the form of `load()` being used
This affects debugging of `.bzl` files that are in the same repository as the target being built.

Let's imagine we have a workspace named `foo` that contains `bar.bzl`, and we want to place a breakpoint in that file. Under Bazel's `execroot`, there will be a symlink `foo` to the actual directory where the workspace exists (let's call it `$HOME/Scratch/foo`), and `bar.bzl` will be in there as expected.

Likewise, files referenced from external repository references are under Bazel's `external` directory (relative to `output_base`). _This includes "external-looking" references to the **same** workspace._ So in our case, if we refer to `@foo//:bar.bzl`, then that shows up as a symlink in `.../external/foo/bar.bzl` that points to the original file.

The problem is, the source location that Bazel tests in the debug server to decide whether a breakpoint should fire differs depending on how the file was `load`ed.

For example, if I do the following:

```
load("@foo//:bar.bzl", ...)
```

Then Bazel asks "is there a breakpoint at `/private/var/tmp/_bazel_allevato/<HASH>/external/foo/bar.bzl`?" This is the form of breakpoint that we send to the debug server [today](https://github.com/bazelbuild/vscode-bazel/blob/master/src/debug-adapter/client.ts#L212-L226). (The reason I implemented this one is because it just happened to work with the repository I was testing against, which references its own `.bzl` files using the repo name.)

However, if I do the following:

```
load("//:bar.bzl", ...)
```

Then Bazel asks "is there a breakpoint at `$HOME/Scratch/foo/bar.bzl`?" In this case, it appears that it's resolving the `execroot/foo` symlink and then looking relative to that location.

Even though these are _the same file_, Bazel's path comparison sees them as being different, and a breakpoint set for one won't fire for the other.

The only way I can see around this for now is for us to send _two_ breakpoints to the server so that the breakpoint fires regardless of whether the `.bzl` file was loaded using a repository-qualified path or not. 😕

allevato:
- Handle file paths correctly for breakpoints.
- Don't cause an exception if the debugger tries to get the frames of an unpaused thread.
- Don't drop events when the sequence number is a Long zero.

Fixes #24.

### DIFF
--- a/src/debug-adapter/client.ts
+++ b/src/debug-adapter/client.ts
@@ -209,10 +209,20 @@ class BazelDebugSession extends DebugSession {
     response: DebugProtocol.SetBreakpointsResponse,
     args: DebugProtocol.SetBreakpointsArguments,
   ) {
-    // VS Code passes us the absolute path to the file, but Bazel's debugger
-    // will match against paths beneath the output_base. It appears that even
-    // .bzl files in the same workspace as the one containing the target being
-    // built will be found in the "external" subtree.
+    // The path we need to pass to Bazel here depends on how the .bzl file has
+    // been loaded. Unfortunately this means we have to create two breakpoints,
+    // one for each possible path, because the way the .bzl file is loaded is
+    // chosen by the user:
+    //
+    // 1. If the file is loaded using an explicit repository reference (i.e.,
+    //    `@foo//:bar.bzl`), then it will appear in the `external` subdirectory
+    //    of Bazel's output_base.
+    // 2. If the file is loaded as a same-repository path (i.e., `//:bar.bzl`),
+    //    then Bazel will treat it as if it were under `execroot`, which is a
+    //    symlink to the actual filesystem location of the file.
+    //
+    // TODO(allevato): We may be able to simplify this once
+    // https://github.com/bazelbuild/bazel/issues/6848 is in a release.
     const workspaceName = path.basename(this.bazelInfo.get("execution_root"));
     const relativeSourcePath = path.relative(
       this.bazelInfo.get("workspace"),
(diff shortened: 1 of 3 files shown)


src/debug-adapter/client.ts
```diff
@@ -209,10 +209,20 @@ class BazelDebugSession extends DebugSession {
...
-    // built will be found in the "external" subtree.
+    // The path we need to pass to Bazel here depends on how the .bzl file has
+    // been loaded. Unfortunately this means we have to create two breakpoints,
+    // one for each possible path, because the way the .bzl file is loaded is
+    // chosen by the user:
+    //
+    // 1. If the file is loaded using an explicit repository reference (i.e.,
+    //    `@foo//:bar.bzl`), then it will appear in the `external` subdirectory
+    //    of Bazel's output_base.
+    // 2. If the file is loaded as a same-repository path (i.e., `//:bar.bzl`),
```
thomasvl: I thought you had said something also about these symlinks not existing until the build is started.  Does that play into anything you have to deal with?

src/debug-adapter/client.ts
```diff
@@ -271,29 +282,32 @@ class BazelDebugSession extends DebugSession {
...
+          vsFrame.line = location.lineNumber;
+        }
+        vsFrames.push(vsFrame);
       }
-      vsFrames.push(vsFrame);
+      response.body = { stackFrames: vsFrames, totalFrames: vsFrames.length };
     }
 
-    response.body = { stackFrames: vsFrames, totalFrames: vsFrames.length };
     this.sendResponse(response);
```
thomasvl: It looks like the response can now get sent without the body having been filled in, is that ok?

allevato: Yeah, they'll just be undefined fields by default, which is fine.

src/debug-adapter/client.ts
allevato: That's only an issue if I need to _resolve_ them, which I'm not doing here; I'm just constructing the paths (or in the case of the new path, using the one I already have).

PR #53, opened 2019-02-01
Add boiler plate setting up integration tests

smukherj1:
Do not merge yet. This commit doesn't really work yet.

The tests run and pass for me when I run them. However, if you look at src/test/extension.test.ts, I'm trying to get the test to fail when I run it to ensure failures will actually be caught. But this isn't happening. Would like some help on figuring out what's wrong with my setup.

### DIFF
--- a/src/buildifier/buildifier.ts
+++ b/src/buildifier/buildifier.ts
@@ -203,10 +203,10 @@ function executeBuildifier(
       [getDefaultBuildifierExecutablePath()].concat(args).join(" "),
       execOptions,
       (error: Error, stdout: string, stderr: string) => {
-        if (error) {
-          reject(error);
-        } else {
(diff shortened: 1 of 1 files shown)


smukherj1:
@achew22 I'm trying to launch the VS Code extension development host and run these tests inside that. Instructions here https://code.visualstudio.com/api/working-with-extensions/testing-extension.

The idea is to get a mock version of the VS Code editor to test parts of this extension that talk to VS Code.

Testing whether the interaction of this extension with Bazel is working correctly is probably something we need to look into as well. However, we aren't using Bazel to build or test this repository so I'm not sure how https://github.com/bazelbuild/bazel-integration-testing  would fit in.

achew22:
Ah, I incorrectly assumed you were using bazel as your build system. If you're not doing that then the integration testing repo probably won't help too much. Good luck with your testing!

PR #59, opened 2019-02-19
Support Bazel's exit codes.

DavidGoldman:


src/bazel/bazel_exit_code.ts
```diff
@@ -0,0 +1,137 @@
...
+  REMOTE_ERROR = "REMOTE_ERROR",
+  LOCAL_ENVIRONMENTAL_ERROR = "LOCAL_ENVIRONMENTAL_ERROR",
+  INTERNAL_ERROR = "INTERNAL_ERROR",
+  PUBLISH_ERROR = "PUBLISH_ERROR",
+  REMOTE_EXECUTOR_OVERLOADED = "REMOTE_EXECUTOR_OVERLOADED",
+  RESERVED = "RESERVED",
+  UNKNOWN = "UNKNOWN",
+}
+
+export function exitCodeToUserString(exitCode: ExitCode): string {
```
allevato: Please add JSDoc blocks for any function (exported or otherwise) that's not a trivial accessor/mutator.

src/bazel/bazel_task_info.ts
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as vscode from "vscode";
+import { IBazelCommandOptions } from "./bazel_command";
+
+/** Information about a Bazel task. */
+export class BazelTaskInfo {
+  /** pid for the task (if started). */
+  private processId: number;
```
allevato: Since this and `exitCode` just have trivial accessors/mutators, you can ditch those methods and just make these public fields (and then update the use sites accordingly). We can still safely update their behavior in the future if we need to by declaring them as properties instead of methods.

src/bazel/bazel_exit_code.ts
DavidGoldman: Done

src/bazel/bazel_task_info.ts
DavidGoldman: Done

PR #77, opened 2019-04-09
Include any targets from the opened directory in the tree view.

### LINKED ISSUE #46: Targets at the top level of a repository don't appear in the tree view
We're not even querying the targets here.

thomasvl:
Fixes #46

### DIFF
--- a/src/workspace-tree/bazel_workspace_folder_tree_item.ts
+++ b/src/workspace-tree/bazel_workspace_folder_tree_item.ts
@@ -33,7 +33,7 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
   }
 
   public getChildren(): Promise<IBazelTreeItem[]> {
-    return this.getPackages();
+    return this.getDirectoryItems();
   }
 
   public getLabel(): string {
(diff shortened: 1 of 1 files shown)


src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -141,7 +143,7 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
   /**
    * Returns a promise for an array of tree items representing build packages.
    */
-  private async getPackages(): Promise<BazelPackageTreeItem[]> {
+  private async getPackages(): Promise<IBazelTreeItem[]> {
```
allevato: Nit: for future us, we should probably change the name of this method now to `getTopLevelItems` (and update the call site in `getChildren`) so it's more accurate.

thomasvl: `getTopLevelItems` isn't really right either, _top_ implies (to me aleast) things at the WORKSPACE level.  This is getting the packages and items from the opened directory.

allevato: Well, the function is kind of odd in the sense that it gets *all* the packages but then builds the whole tree and only returns the top-level items.

`buildPackageTreeAndReturnTopLevelPackagesAndTargets`? 😅

thomasvl: That better?

allevato: WFM

Issue #72, opened 2019-03-19
Build all targets inside a package


A-Sam:
ATM, all targets are being loaded in the view (Bazel build targets). Every time i want to run all the targets inside a package i have to write the 1 liner for loop in bash to build all targets read from $(bazel query //...). It would be good option to add to this view.

thomasvl:
fyi - `bazel build //...`, no need for the query; you can also do `bazel build //package:all` to build everything in the package, but not do it recursively.

A-Sam:
Yeah. I use the query though to see the targets prior building and saving them to a file(smth like a dry-run). But I think it would be nice to add a feature on Bazel Build Targets to run all targets under a specific package in the extension.

thomasvl:
Now that we have the right click on the packages in the tree view, we could also change the treeview code to always show the opened directory node even when only one directory is open in VS Code, that catch is that directory might not have a BUILD file (it could be some random place within the overall bazel workspace).  So building `:all` there would error, but `...` would work, so the code would need to know to only enable some vs. all of the commands.

A-Sam:
Bazel build //target/... will build all targets in the package which is what i originally meant and use.. thumbs up! Will this be released any time soon?

thomasvl:
The 0.2.0 release in May should have had it.

allevato:
Closing because this was released in 0.2.0.

oliverdain:
It does not seem to be possible to build `//...`. If I select `Bazel build package recursively` I have to select a package and `//` is not an option. I have tons of packages so it'd be really nice to be able to just build all of them.

thomasvl:
Please open a new issue specific to the whole WORKSPACE cases.

PR #81, opened 2019-04-09
Add support for build/testing packages.

thomasvl:
- Add context command to the treeview on the packages to do build and testing
  of the package or the package resursively (:all vs /...).
- Includes a quickpick for generally using the command and selecting a package.

Mostly addresses #72, the limitation being the current UI doesn't really every
provide the "root" (the directory opened in VS Code) as a package, it just shows
the sub packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,8 +40,28 @@
             },
             {
                 "category": "Bazel",
+                "command": "bazel.buildAll",
+                "title": "Build Package"
+            },
+            {
+                "category": "Bazel",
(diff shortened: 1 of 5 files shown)


thomasvl:
>IIRC the logic is that we show the roots if there are multiple directories open in the VS Code workspace. We could just force that to show it all the time.
>
> Do you want to do that now, or separately? Up to you.

The other issue is still open, so we can use that to track it for the moment.  This change is already a little big, so making the change in another PR seems better from a history pov.

PR #82, opened 2019-04-09
Improvements to package queries.

### LINKED ISSUE #79: The query for collecting packages only catches ones with rules also
To build the tree view, we do `bazel query --output=package ...`, but all it really takes to make a package is for there to be an empty `BUILD` file, but since that empty file doesn't have any targets declared, the query doesn't return it, so the tree view doesn't make a node there.  You can see this with the `examples` directory in https://github.com/bazelbuild/rules_apple.

thomasvl:
- Always sort the result (callers were).
- Tweak the query so packages that are created by an empty BUILD file still
  get captured (done by using a query that catches the BUILD file target
  itself (it is exposed as a source file))

Fixes #79

### DIFF
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -97,10 +97,9 @@ async function queryWorkspaceQuickPickPackages(
 ): Promise<BazelTargetQuickPick[]> {
   const packagePaths = await new BazelQuery(
     workspaceInfo.bazelWorkspacePath,
-    "...",
-    ["--output=package"],
+    "...:*",
+    [],
   ).queryPackages();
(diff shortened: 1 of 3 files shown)


src/bazel/bazel_quickpick.ts
```diff
@@ -97,10 +97,9 @@ async function queryWorkspaceQuickPickPackages(
 ): Promise<BazelTargetQuickPick[]> {
   const packagePaths = await new BazelQuery(
     workspaceInfo.bazelWorkspacePath,
-    "...",
-    ["--output=package"],
+    "...:*",
```
allevato: Nice! I didn't realize this would do that.

PR #90, opened 2019-04-12
Add logging output channel

reinhillmann:
Creates a logging output channel called "VSCode-Bazel". Components can then log to this channel to show debug information.

Adds a new command: "Show Extension Log" which will display the channel.

Also adds a new settings: "bazel. showExtensionLog" which will automatically display the output channel when the plugin initializes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,11 @@
                     "dark": "./icons/refresh-dark.svg",
                     "light": "./icons/refresh-light.svg"
                 }
+            },
+            {
+                "category": "Bazel",
+                "command": "bazel.showExtensionLog",
+                "title": "Show Extension Log"
             }
         ],
         "configuration": {
@@ -96,6 +101,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Whether to automatically apply lint fixes from buildifier when formatting a Bazel file."
+                },
+                "bazel.showExtensionLog": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Show the VSCode-Bazel output channel when the Bazel extension initializes."
                 }
             }
         },
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -15,7 +15,6 @@
 import * as vscode from "vscode";
 import * as which from "which";
 
-import { build } from "protobufjs";
 import {
   BazelWorkspaceInfo,
   createBazelTask,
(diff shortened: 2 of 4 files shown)


src/extension/extension.ts
```diff
@@ -37,13 +35,20 @@ import { BazelBuildCodeLensProvider } from "../codelens";
 import { BazelTargetSymbolProvider } from "../symbols";
 import { BazelWorkspaceTreeProvider } from "../workspace-tree";
 
+/**
+ * Allows logging to the custom output channel "VSCode-Bazel".
+ */
+export let logOutputChannel: vscode.OutputChannel | undefined;
```
allevato: This would require other components to import `extension/extension` in order to access this channel, which seems like the wrong dependency direction (and could lead to cycles). `extension/extension` is essentially the "main" file that shouldn't be imported by anyone else.

Instead, an approach like this might work: create a new `src/logging/logging.ts` file (along with the `index.ts` following the pattern of the other directories) and move the channel variable and setup function in there. Then, also create and export a log method (`bazelLog`?) that logs to the channel declared in that file. This lets you keep the channel itself private and only export the setup and logging functions, which clients would call.

reinhillmann: Done.

reinhillmann:
> > Creates a logging output channel called "VSCode-Bazel". Components can then log to this channel to show debug information.
> 
> What are the "components" in this case, and what kind of information will they be logging?

Components would be, for example, the bazel_query or buildifier code components. They would be logging successes and errors instead of sending them to the Visual Studio alert toaster popups. These error messages are not easily copied from the alert popups.

package.json
```diff
@@ -76,6 +76,11 @@
                     "dark": "./icons/refresh-dark.svg",
                     "light": "./icons/refresh-light.svg"
                 }
+            },
+            {
+                "category": "Bazel",
+                "command": "bazel.showOutput",
+                "title": "Show Output"
```
allevato: I think a more descriptive title would be helpful here, like "Show Extension Log". "Show Output" sounds to me like Bazel stdout/stderr output.

Likewise, can you make the command name `bazel.showExtensionLog`?

reinhillmann: Done.

reinhillmann:
> "Instead of" or "In addition to"? I can see the value of a log window that keeps track of session errors and makes them easy to copy, but I would prefer that the popups remain as well; they draw the user's attention to the error better than a log window (which may not always be visible).

Yes, in addition to. The existing toasts remain.

Issue #88, opened 2019-04-12
getDefaultBazelExecutablePath() returns empty string


ashi009:
I have an empty setting for `bazel.executable` and the build task failed due to "build: command not found". After manually set `bazel.executable` to "bazel", build task works. I believe that `getDefaultBazelExecutablePath()` returns an empty string for some reason (though the code looks sane to me).

Please consider setting "bazel" as the default for `bazel.executable` or using [`WorkspaceConfiguration.get<T>(section: string, defaultValue: T)`](https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration) instead.

This is also the case for `bazel.buildifierExecutable`.

![image](https://user-images.githubusercontent.com/687367/56042739-5415a080-5d6e-11e9-8f2f-50b4f8f4aecd.png)

![image](https://user-images.githubusercontent.com/687367/56042715-4829de80-5d6e-11e9-9b84-bbee733835fb.png)

thomasvl:
The code tries to handle this already, for the 0.1.0 release it should be hitting: http://go/gh/bazelbuild/vscode-bazel/blob/0.1.0/src/bazel/bazel_command.ts#L96

Can you trace through there, what are you getting back that the check for a zero length string isn't working?

allevato:
Was this ever resolved?

PR #92, opened 2019-04-15
Remove vscode dependency from bazel_command

reinhillmann:
bazel_command currently has a dependency on vscode for the Bazel executable path. This change is to remove that dependency by passing the executable path into the constructor of the BazelCommand class.

This has the following benefits:
1. This will allow us to eventually move the Bazel-specific logic into a separate library.
2. It makes the code more testable.

### DIFF
--- a/src/extension/configuration.ts
+++ b/src/extension/configuration.ts
@@ -1,0 +1,34 @@
+// Copyright 2018 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


src/extension/configuration.ts
```diff
@@ -0,0 +1,34 @@
...
+ * configuration, or just "bazel" if not present (in which case the system path
+ * will be searched).
+ */
+export function getDefaultBazelExecutablePath(): string {
+  // Try to retrieve the executable from VS Code's settings. If it's not set,
+  // just use "bazel" as the default and get it from the system PATH.
+  const bazelConfig = vscode.workspace.getConfiguration("bazel");
+  const bazelExecutable = bazelConfig.executable as string;
+  if (bazelExecutable.length === 0) {
+    return "bazel";
```
thomasvl: I guess this also fixes #88?  Not sure if the suggested api there is any better.

reinhillmann: I didn't make any changes to this function, I just moved it around. So if it does fix it then it was fixed before. Since this bug also references buildifier I think we should leave it open until that is also addressed.

allevato: Yeah, I'm not sure how #88 is happening since we explicitly have code that avoids the empty string case. It's worth looking into, but separately.

allevato:
@reinhillmann It looks like this needs to be rebased on top of `master` before I can merge it.

PR #95, opened 2019-04-16
Use `npm clean-install` for CI.

thomasvl:
Looking at `npm help clean-install` it seems like CI setups are better off
using it (vs. `install`) because it won't try to update the package json
files.


thomasvl:
Hm, looking at the failure, I guess the support for `npm run check-lint` isn't there if we just do a `clean-install`?

thomasvl:
`ci` works, I guess `clean-install` is _newer_ that then `npm` on the BazelCI machines.  Anyway, squashed/updated, once it works again, should be good to go.

PR #99, opened 2019-04-16
Fix error when user has multiple workspace folders open

reinhillmann:
We try to create a `BazelWorkspaceFolderTreeItem` for every folder the user has opened regardless of whether those folders are Bazel WORKSPACE folders.
This fix will ignore non-Bazel folders when creating
 `BazelWorkspaceFolderTreeItem`s.

### DIFF
--- a/src/workspace-tree/bazel_workspace_tree_provider.ts
+++ b/src/workspace-tree/bazel_workspace_tree_provider.ts
@@ -83,11 +83,17 @@ export class BazelWorkspaceTreeProvider
       // If the user has multiple workspace folders open, then show them as
       // individual top level items.
       return Promise.resolve(
-        vscode.workspace.workspaceFolders.map((folder) => {
-          return new BazelWorkspaceFolderTreeItem(
-            BazelWorkspaceInfo.fromWorkspaceFolder(folder),
-          );
-        }),
+        vscode.workspace.workspaceFolders
+          .map((folder) => {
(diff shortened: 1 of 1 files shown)


src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -84,9 +84,11 @@ export class BazelWorkspaceTreeProvider
...
       return Promise.resolve(
         vscode.workspace.workspaceFolders.map((folder) => {
-          return new BazelWorkspaceFolderTreeItem(
-            BazelWorkspaceInfo.fromWorkspaceFolder(folder),
-          );
+          const workspaceInfo = BazelWorkspaceInfo.fromWorkspaceFolder(folder);
+          if (workspaceInfo) {
+            return new BazelWorkspaceFolderTreeItem(workspaceInfo);
+          }
+          return undefined;
```
allevato: This will put `undefined` elements in the array that the promise resolves to, which might end up being harmless but we shouldn't depend on them being ignored.

Since `flatMap` isn't well supported yet apparently (it's new in ES2019), you'll need to chain a `filter` after the `map` call:

```typescript
vscode.workspace.workspaceFolders.map((folder) => {
  ...
}).filter((folder) => { folder !== undefined }),
```

reinhillmann: It does get ignored but I agree that it feels a little hopeful without the filter. Added.

PR #102, opened 2019-04-16
Move Bazel workspace to bazellib

reinhillmann:
As part of this change:
1. Move the `BazelWorkspaceInfo` class to bazellib.
2. Remove the VSCode `workspaceFolder` object from `BazelWorkspaceInfo` and opt to pass it explicitly when constructing a `BazelWorkspaceFolderTreeItem`. This removes the tight coupling between a VSCode `workspaceFolder` and a Bazel workspace.
3. Simplify the `BazelWorkspaceInfo` as there is no longer need to differentiate between file or folder paths.
4. Changed the task scope to global as per the todo message.


src/bazel/tasks.ts
```diff
@@ -56,9 +56,7 @@ export function createBazelTask(
   const targetsDescription = options.targets.join(", ");
   const task = new vscode.Task(
     { type: "bazel", command, targets: options.targets },
-    // TODO(allevato): Change Workspace to Global once the fix for
-    // Microsoft/vscode#63951 is in a stable release.
-    options.workspaceInfo.workspaceFolder || vscode.TaskScope.Workspace,
```
allevato: There's a subtle difference here that's being lost. If the command options referred to a workspace folder, the created task would be scoped to that workspace folder; _otherwise_ the task would be created at `Workspace` scope (which was a substitute for `Global` scope due to that bug).

This change makes _all_ tasks globally scoped, which could cause different behaviors depending on what other configurations those workspace folders might have compared to the global (user) configuration.

In order to preserve the old behavior, you'll probably need to map the file system path of the `workspaceInfo` to the `vscode.workspace.WorkspaceFolder` value it corresponds to, since you no longer carry that information directly.

reinhillmann: Ah that makes more sense. I've updated the change so that tasks can accept the VSCode workspaceFolder and set the scope accordingly. I also pass that workspace folder whenever I create a task. PTAL.

reinhillmann: Please don't approve yet. There's more work to be done since there won't always be an active document.

allevato: Sounds good, let me know when it's ready.

reinhillmann:
Closing this for now. May revisit later.

Issue #110, opened 2019-05-19
Add a "Copy to Clipboard" context menu to the Bazel Workspace Tree view


VanBronckhorst:
It would be useful to provide a copy to clipboard option in the Bazel workspace Tree view. 
This would copy the target name so it can easily be added to the list of dependencies in a BUILD file.

PR #112, opened 2019-05-19
Add a context menu option to the bazel workspace tree that allows the…

VanBronckhorst:
#110 
image


VanBronckhorst:
Another fail rebase :/

PR #111, opened 2019-05-19
Add "Copy to Clipboard" context menu option

VanBronckhorst:
#110 
image

PR #113, opened 2019-05-19
Add "Copy to Clipboard" context menu option

### LINKED ISSUE #110: Add a "Copy to Clipboard" context menu to the Bazel Workspace Tree view

VanBronckhorst:
Fixed #110.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-bazel",
     "displayName": "vscode-bazel",
     "description": "Bazel BUILD integration",
-    "version": "0.1.0",
+    "version": "0.2.0",
     "publisher": "BazelBuild",
     "license": "Apache-2.0",
     "repository": {
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     },
     "icon": "media/bazel-logo.png",
     "engines": {
-        "vscode": "^1.27.0"
+        "vscode": "^1.30.0"
     },
     "categories": [
         "Programming Languages"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     },
     "icon": "media/bazel-logo.png",
     "engines": {
-        "vscode": "^1.27.0"
+        "vscode": "^1.30.0"
```
allevato: Is this change necessary for any of the functionality in this PR? If not, please revert it and also revert the changes to package-lock.json.

package.json
```diff
@@ -81,6 +81,11 @@
                 "category": "Bazel",
                 "command": "bazel.showExtensionLog",
                 "title": "Show Extension Log"
+            },
+            {
+                "category": "Bazel",
+                "command": "bazel.copyTargetToClipboard",
+                "title": "Copy to Clipboard"
```
allevato: Let's call this "Copy Label to Clipboard" to make it clear what's being copied (the [label](https://docs.bazel.build/versions/master/build-ref.html#labels) as opposed to, say, file contents of some sort).

package.json
allevato: A colleague pointed out that 1.30 is when the clipboard APIs were introduced. I've pushed a separate change to update the version requirement (#114); can you revert these package(-lock).json changes and rebase so this will be a standalone change?

VanBronckhorst: Yeah, that's why I changed the min-version, thanks for your change.

package.json
VanBronckhorst: Ah thanks, I did not know the //path was called label.

VanBronckhorst:
BTW this solves #110 , not #111  as per first comment

PR #117, opened 2019-05-31
Start using buildifier's recent JSON output.

allevato:
This was added in 0.25.1; the extension now checks if the version is compatible and directs the user to download a newer one if necessary.

This gives us more refined diagnostics. For example, we can now highlight the specific part of the code where the warning occurs instead of the entire line:

screenshot


src/buildifier/buildifier_availability.ts
```diff
@@ -0,0 +1,87 @@
...
+          try {
+            JSON.parse(stdout);
+            return;
+          } catch {
+            // Swallow the error; we'll display the prompt below.
+          }
+        }
+        // If we didn't get valid JSON back, we don't have a compatible version.
+        showBuildifierDownloadPrompt(
+          "Buildifier is too old (0.25.1 or higher is needed)",
```
laurentlb: Can this happen for another reason?
We could include the error output to help diagnose the problem.

allevato: This would happen if:

1) buildifier's invocation changed in such a way that the flags we pass in are no longer valid
2) buildifier stops treating the empty file as valid

I'm not sure where we could include the whole error output that wouldn't overpopulate the UI; for example, in the case of a too-old buildifier, the output is the entire usage/help page.

allevato: I'm going to go ahead and merge this, but if I think of a good way to surface the full error output, I'll circle back around to it.

PR #121, opened 2019-06-19
Add "starlark" language to formatting provider.

dierksen:
If users have other starlark-ish files to format that don't meet the patterns here, there's no way to take advantage of the formatter. This enables using a file association in the workspace/user settings to format anything set as "starlark".

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -88,6 +88,7 @@ export function activate(context: vscode.ExtensionContext) {
     // Buildifier formatting support
     vscode.languages.registerDocumentFormattingEditProvider(
       [
+        { language: "starlark"},
         { pattern: "**/BUILD" },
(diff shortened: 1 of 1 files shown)


src/extension/extension.ts
```diff
@@ -88,6 +88,7 @@ export function activate(context: vscode.ExtensionContext) {
     // Buildifier formatting support
     vscode.languages.registerDocumentFormattingEditProvider(
       [
+        { language: "starlark"},
```
allevato: Super-nit: Please add space before the closing brace (why is tslint not complaining?)

dierksen: User error on my end: I had my global tslint misconfigured so it wasn't using the right setup. Fixed.

allevato: It should have complained on our CI too, though. I'll have to look into it.

PR #126, opened 2019-09-09
Fix buildifier file type detection

whilp:
Adjust code to match original intent. `pathObject` properties `base` and `ext` have values like "BUILD.bazel" and ".bazel", respectively. For the comparisons here, we want the results of `path.basename()` with its `ext` parameter set.

Fixes https://github.com/bazelbuild/vscode-bazel/issues/125.

### DIFF
--- a/src/buildifier/buildifier.ts
+++ b/src/buildifier/buildifier.ts
@@ -126,14 +126,26 @@ export function getBuildifierFileType(fsPath: string): BuildifierFileType {
   // https://github.com/bazelbuild/buildtools/blob/d39e4d/build/lex.go#L88)
   // so that user actions in the IDE are consistent with the behavior they
   // would see running buildifier on the command line.
-  const parsedPath = path.parse(fsPath.toLowerCase());
+  const raw = fsPath.toLowerCase();
+  let parsedPath = path.parse(raw);
+  if (parsedPath.ext === ".oss") {
+    parsedPath = path.parse(parsedPath.name);
+  }
   if (parsedPath.ext === ".bzl" || parsedPath.ext === ".sky") {
     return "bzl";
   }
-  if (parsedPath.ext === ".build" || parsedPath.base === "build") {
+  if (
+    parsedPath.ext === ".build" ||
+    parsedPath.name === "build" ||
+    parsedPath.name.startsWith("build.")
+  ) {
     return "build";
   }
-  if (parsedPath.ext === ".workspace" || parsedPath.base === "workspace") {
+  if (
+    parsedPath.ext === ".workspace" ||
+    parsedPath.name === "workspace" ||
+    parsedPath.name.startsWith("workspace.")
+  ) {
     return "workspace";
   }
   return "bzl";


src/buildifier/buildifier.ts
```diff
@@ -126,14 +126,16 @@ export function getBuildifierFileType(fsPath: string): BuildifierFileType {
...
-  const parsedPath = path.parse(fsPath.toLowerCase());
-  if (parsedPath.ext === ".bzl" || parsedPath.ext === ".sky") {
+  const ext = path.extname(fsPath);
+  const base = path.basename(fsPath, ext);
+
+  if (ext === ".bzl" || ext === ".sky") {
     return "bzl";
   }
-  if (parsedPath.ext === ".build" || parsedPath.base === "build") {
+  if (ext === ".build" || base === "build") {
```
allevato: I don't think this is quite what we want to do. The use of `parsedPath.base` here was intentional—it was to say "if the file's extension is `.build` or the _file's entire name, lacking an extension,_ is `BUILD`", then treat it as a "build" file.

You're right that that doesn't account for `BUILD.bazel`, but allowing anything with a basename (sans extension) of `build` is too permissive.

Digging deeper, according to [this bit of code in Bazel](https://source.bazel.build/intellij/+/master:base/src/com/google/idea/blaze/base/bazel/BazelBuildSystemProvider.java;l=79), the only filenames that it will ever accept for BUILD files are `BUILD` and `BUILD.bazel`. So it's probably sufficient to just explicitly look for those two and not worry about anything else.

That won't handle some edge cases, like the [`BUILD.overlay` files](https://github.com/bazelbuild/rules_swift/blob/master/third_party/com_github_apple_swift_protobuf/BUILD.overlay) I use in the rules_swift repository, but since that's a me-convention and not a Bazel-convention, I don't think it's worth the extension trying to support that. If we _really_ wanted to, we could add a configuration setting that would let folks enter additional file patterns that they'd like to treat as build files, but I don't think it's a terribly important addition.

Likewise, the only permitted filename for the workspace file appears to be `WORKSPACE`, so we shouldn't even bother checking `".workspace"`. Honestly, I'm not sure why we have that 🤷‍♂ 

So my recommendation would be to go back to using `path.parse` (and to keep the lowercasing, to handle case-insensitive file systems), then to just check `parsedPath.base === "build" || parsedPath.base === "build.bazel"` here, and only check for `"workspace"` below.

How does that sound?

whilp: Thanks for the quick feedback and for finding the bazel precedents. My change was motivated by:

a) my (selfish) usecase, in which I want the standardish `BUILD.bazel` files to be recognized as build files, and
b) the precedent in `buildifier`.

I arrived at (b) from the comments in the code here. It seems that `vscode-bazel` has looked to `buildifier` to inform its type detection (and will, at some point, just set `--path` when invoking `buildifier`; for what it's worth, I believe that would also fix my problem).

`buildifier` has recently liberalized its file detection in https://github.com/bazelbuild/buildtools/pull/712:

https://github.com/bazelbuild/buildtools/blob/f8b2ce4ca484e24bddd22d1eca9c5746cfd4bf4c/build/lex_test.go#L23-L58

I /think/ it's reasonable for a user to expect `vscode-bazel`'s `buildifier` mode to work exactly as `buildifier` does when invoked directly. If you agree with that general direction, I can take another stab here.

Here's the current full logic in `buildifier`:

https://github.com/bazelbuild/buildtools/blob/f8b2ce4ca484e24bddd22d1eca9c5746cfd4bf4c/build/lex.go#L108-L131

> ...keep the lowercasing...

Buh, that was totally an oversight on my part, thanks for catching. :/ willfix

allevato: Ok, I'm convinced—you fairly point out that since we're invoking _buildifier_, it makes sense to enforce what they handle. I think it raises more questions about why buildifier and Bazel enforce different rules for what it recognizes as BUILD and WORKSPACE files, but that's not a question for vscode-bazel. 🙂 

And taking a closer look at the logic in the original Go code, it looks like their notion of `base` is _without_ the extension, so your change _was_ closer to what it's expecting.

I retract my concerns 😄   While you're in here, would you mind adding the bit that removes the `.oss` extension if it's present? This is a good opportunity to just bring the whole function up to date with the current state of things.

FWIW I think it's still fine to continue to use `path.parse` in the implementation, unless you think there's a reason to switch to the more primitive methods. If we want the base without the extension, we just need to use `parsedPath.name` instead of `parsedPath.base`.

laurentlb: Regarding Buildifier: it's useful to format/lint files like `BUILD.oss` (which becomes a BUILD file when exported) or [`six.BUILD`](https://github.com/tensorflow/tensorflow/tree/master/third_party) (it becomes a BUILD file when imported from a workspace).

whilp: @allevato thanks again for the feedback. I pushed an update that should do exactly what `buildifier` does; I verified that by copying the buildifier test case:

https://github.com/bazelbuild/vscode-bazel/compare/master...whilp:whilp/test-buildifier-filetype?expand=1

src/buildifier/buildifier.ts
```diff
@@ -126,14 +126,26 @@ export function getBuildifierFileType(fsPath: string): BuildifierFileType {
...
+    parsedPath = path.parse(parsedPath.name);
+  }
   if (parsedPath.ext === ".bzl" || parsedPath.ext === ".sky") {
     return "bzl";
   }
-  if (parsedPath.ext === ".build" || parsedPath.base === "build") {
+  if (
+    parsedPath.ext === ".build" ||
+    parsedPath.name === "build" ||
+    parsedPath.name.startsWith("build")
```
allevato: You've dropped the dot from the end of this check (and below, with `workspace`). The original Go check is [`strings.HasPrefix(base, "build.")`](https://github.com/bazelbuild/buildtools/blob/master/build/lex.go#L125); presumably this lets them catch files named like "BUILD.foo.bar" where stripping the extension only leaves "BUILD.foo".

Without the dot, this will also match things like "buildfoo", which could lead to subtle differences.

whilp: Oh yep! Thanks for your patience, fixin...

PR #138, opened 2019-10-24
Add a setting to support disabling the building support.

thomasvl:
Mainly useful as a global setting (not per project), lets you disable some
WORKSPACE discovery/target treeview where but still support the other features
(buildifer, BUILD syntax support, etc.).

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
                 },
                 "bazel.directoriesToIgnore": {
                     "type": "string",
+                    "scope": "application",
                     "default": "",
                     "description": "Regex of directors to ignore when trying to figure out if something is in a Bazel Workspace. This just impacts things that need a Bazel Workspace, BUILD/.bzl syntax and buildifier support still work."
                 }
--- a/package.json
+++ b/package.json
@@ -111,6 +111,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Show the VSCode-Bazel output channel when the Bazel extension initializes."
+                },
+                "bazel.directoriesToIgnore": {
+                    "type": "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -111,6 +111,11 @@
                     "type": "boolean",
                     "default": false,
                     "description": "Show the VSCode-Bazel output channel when the Bazel extension initializes."
+                },
+                "bazel.directoriesToIgnore": {
+                    "type": "string",
+                    "default": "",
+                    "description": "Regex of directors to ignore when trying to figure out if something is in a Bazel Workspace. This just impacts things that need a Bazel Workspace, BUILD/.bzl syntax and buildifier support still work."
```
allevato: It looks like the logic below matches the entire path, not just "directories", right? Should this just be called "pathsToIgnore" then, with the following suggested description?

"Regex of paths to ignore when trying to figure out if a file is in a Bazel workspace. This impacts operations that require a workspace, like queries, but syntax highlighting and buildifier work regardless."

src/bazel/bazellib/bazel_utils.ts
```diff
@@ -51,6 +52,31 @@ export async function getTargetsForBuildFile(
   return queryResult;
 }
 
+/**
+ * Check if a path should be ignored and not considered to be part of a
+ * Bazel Workspace.
+ *
+ * @param fsPath The path to a file in a Bazel workspace.
+ * @returns true / false for if the path should be ignore(assumed not to
```
allevato: Nit: Space before open paren

package.json
thomasvl: done

PR #140, opened 2019-10-25
Change pathsToIgnore to be a list.

thomasvl:
Should make it easier if needing more than one thing.

The VSCode docs didn't seem to document this, but the UI supports it and other
extensions use it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,10 +113,14 @@
                     "description": "Show the VSCode-Bazel output channel when the Bazel extension initializes."
                 },
                 "bazel.pathsToIgnore": {
-                    "type": "string",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "uniqueItems": true,
                     "scope": "application",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -113,10 +113,14 @@
...
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "uniqueItems": true,
                     "scope": "application",
-                    "default": "",
-                    "description": "Regex of paths to ignore when trying to figure out if a file is in a Bazel workspace. This impacts operations that require a workspace, like queries, but syntax highlighting and buildifier work regardless."
+                    "default": [],
+                    "description": "A list of Regexs that when matched cause paths not to be check for being in a WORKSPACE. This impacts operations that require a workspace, like queries, but syntax highlighting and buildifier work regardless."
```
allevato: s/check/checked/

thomasvl: done.

PR #143, opened 2019-11-08
Add settings for startup options and command args.

### LINKED ISSUE #134: Allow setting build config (--config=...) when building targets directly from BUILD file
Hello,

we require setting some config for the bazel build:

`bazel build --config=xyz //path/to/target:target`

I just tried using the vscode-bazel plugin and building targets directly from the BUILD file. Unfortunately this will fail for us since I can't see any way to set the --config option.

Is there a way to set --config?

thomasvl:
They follow the normal vscode support, they could be per directory, workspace,
or user.

Fixes #134

### DIFF
--- a/src/bazel/tasks.ts
+++ b/src/bazel/tasks.ts
@@ -30,19 +30,16 @@ function quotedOption(option: string): vscode.ShellQuotedString {
  *
  * @param command The Bazel command to execute.
  * @param options Describes the options used to launch Bazel.
- * @param addDefaults If the user's defaults startup and command args should be
- *     added.
  */
 export function createBazelTask(
   command: "build" | "clean" | "test",
   options: IBazelCommandOptions,
-  addDefaults = true,
 ): vscode.Task {
   const bazelConfigCmdLine = vscode.workspace.getConfiguration("bazel.commandLine");
-  const startupOptions: [string] =
-    (addDefaults ? bazelConfigCmdLine.startupOptions : []);
+  const startupOptions: [string] = bazelConfigCmdLine.startupOptions;
+  const addCommandArgs = command === "build" || command === "test";
   const commandArgs: [string] =
-    (addDefaults ? bazelConfigCmdLine.commandArgs : []);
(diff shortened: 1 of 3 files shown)


src/bazel/tasks.ts
```diff
@@ -30,12 +30,23 @@ function quotedOption(option: string): vscode.ShellQuotedString {
  *
  * @param command The Bazel command to execute.
  * @param options Describes the options used to launch Bazel.
+ * @param addDefaults If the user's defaults startup and command args should be
+ *     added.
  */
 export function createBazelTask(
   command: "build" | "clean" | "test",
   options: IBazelCommandOptions,
+  addDefaults = true,
```
allevato: Do you think we'll want to toggle this separately for build/test elsewhere, or should we just do this if `command == "clean"` and drop the argument?

thomasvl: The api was generic enough, I wasn't sure if it could ever be used any place else.  Anyway, up to you.

src/extension/extension.ts
```diff
@@ -406,7 +406,7 @@ function onTaskProcessEnd(event: vscode.TaskProcessEndEvent) {
       const timeInSeconds = measurePerformance(bazelTaskInfo.startTime);
       vscode.window.showInformationMessage(
         `Bazel ${
-          bazelTaskInfo.command
```
allevato: I wonder why this changed. Different versions of prettier or tslint?

src/bazel/tasks.ts
allevato: Let's condition it on the command for now. With this API, it's possible for the user to forget to pass the argument and then have build/test flags passed to `clean`, which is odd.

Actually, looking closer, don't we still need to support startup options for `clean`? One of the options could be `--output_base`, which would change where Bazel actually expects to clean.

thomasvl: Done

src/extension/extension.ts
thomasvl: no clue.  fixed.

PR #158, opened 2019-12-09
Improve direct process handing

thomasvl:


src/bazel/bazel_command.ts
```diff
@@ -65,17 +65,12 @@ export abstract class BazelCommand {
...
-      result += " ";
-      result += this.options.join(" ");
-    }
-    if (additionalOptions.length > 0) {
-      result += " ";
-      result += additionalOptions.join(" ");
-    }
+  /** The args used to execute the for the command. */
+  protected execArgs(additionalOptions: string[] = []) {
+    const result = [`${this.bazelCommand()}`]
```
allevato: You can just use `[this.bazelCommand()]` here without the interpolation.

src/buildifier/buildifier_availability.ts
```diff
@@ -40,8 +40,9 @@ export function checkBuildifierIsAvailable() {
     // If we found it, make sure it's a compatible version by running
     // buildifier on an empty input and see if it exits successfully and the
     // output parses.
-    const process = child_process.exec(
-      `${buildifierExecutable} --format=json --mode=check`,
+    const process = child_process.execFile(
+      `${buildifierExecutable}`,
```
allevato: `buildifierExecutable` without interpolation.

src/buildifier/buildifier_availability.ts
```diff
@@ -40,8 +40,9 @@ export function checkBuildifierIsAvailable() {
...
+      [`--format=json`, `--mode=check`],
```
allevato: Use double-quotes instead of backticks around these now that they don't need to be interpolated (I don't know if there's a performance implication, but it's clearer).

src/bazel/bazel_command.ts
thomasvl: done

src/buildifier/buildifier_availability.ts
thomasvl: done

src/buildifier/buildifier_availability.ts
thomasvl: done

Issue #75, opened 2019-04-02
Interferes with ibazel 


alexeagle:
Using the vscode plugin, every time a file is saved it seems to do same `bazel query` calls to the local bazel server.

If you're also running a watch mode at the same time, it takes a long time to get a result in the ibazel session because the server is locked:

```
Build graph changed: "/usr/local/google/home/alexeagle/Projects/angular-bazel-example/src/lib/shorten/BUILD.bazel". Requerying...
Querying for files to watch...
Starting local Bazel server and connecting to it...
Loading: 0 packages loaded
Loading: 9 packages loaded
    currently loading: @build_bazel_rules_nodejs//internal/web_package ... (12 packages)
Loading: 375 packages loaded

Loading: 375 packages loaded
Another command holds the client lock: 
pid=147569
owner=client
cwd=/usr/local/google/home/alexeagle/Projects/angular-bazel-example

Waiting for it to complete...
Another command holds the client lock: 
pid=147575
owner=client
cwd=/usr/local/google/home/alexeagle/Projects/angular-bazel-example

Waiting for it to complete...
Starting local Bazel server and connecting to it...
Loading: 0 packages loaded
Loading: 0 packages loaded
Loading: 375 packages loaded

Loading: 375 packages loaded
Runing //src:devserver
Notifying of changes
Loading: 
Loading: 0 packages loaded
Analyzing: target //src:devserver (0 packages loaded, 0 targets configured)
Analyzing: target //src:devserver (5 packages loaded, 26419 targets configured)
INFO: Analysed target //src:devserver (5 packages loaded, 26424 targets configured).
INFO: Found 1 target...
[0 / 3] [-----] BazelWorkspaceStatusAction stable-status.txt
[27 / 40] checking cached actions
[43 / 48] Compiling TypeScript (devmode) //src/lib/shorten:shorten; 0s worker
[45 / 48] Compiling Angular templates (ngc) //src/app/hello-world:hello-world; 0s worker
[45 / 48] Compiling Angular templates (ngc) //src/app/hello-world:hello-world; 1s worker
[46 / 48] Compiling Angular templates (ngc) //src/app:app; 0s worker
Target //src:devserver up-to-date:
  dist/bin/src/devserver_launcher.sh
  dist/bin/src/devserver
INFO: Elapsed time: 10.939s, Critical Path: 6.00s
INFO: 4 processes: 4 worker.
INFO: Build completed successfully, 9 total actions
[...]

PR #160, opened 2019-12-10
Make queries use --output_base.

thomasvl:
By default, move all queries over to a custom --output_base. This keep the
bazel lock from interfering with user done builds/ibazel/etc.

Helps with #75.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,11 @@
                     "default": false,
                     "description": "Whether to automatically apply lint fixes from buildifier when formatting a Bazel file."
                 },
+                "bazel.queriesShareServer": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Use same Bazel server for queries and builds."
+                },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -102,6 +102,11 @@
                     "default": false,
                     "description": "Whether to automatically apply lint fixes from buildifier when formatting a Bazel file."
                 },
+                "bazel.queriesShareServer": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Use same Bazel server for queries and builds."
```
allevato:
```suggestion
                    "description": "Use the same Bazel server for queries and builds. By default, vscode-bazel uses a separate server for queries so that they can be executed in parallel with builds. You can enable this setting if running multiple Bazel servers has a negative performance impact on your system, but you may experience degraded performance in Visual Studio Code for operations that require queries."
```

Issue #54, opened 2019-02-05
Syntax highlighting in BUILD files


laurentlb:
I don't get syntax highlighting only in `BUILD` files (language mode defaults to "Plain text").
If I rename my file to `BUILD.bazel`, the syntax highlighting turns on.

allevato:
`BUILD` files should be recognized since they're listed in the package.json file here: https://github.com/bazelbuild/vscode-bazel/blob/master/package.json#L152

If you invoke the "Developer: Toggle Developer Tools" command, do you see anything in the console log that would indicate what might be happening?

laurentlb:
No - no message in the console.

Features like "Format document" and the target overlay are working.

allevato:
In the bottom-right area of the status bar, there should be a language identifier. Does it say "Starlark" or "Plain Text"? And if it's the latter, does changing it to "Starlark" fix it?

laurentlb:
It says "Plain Text". Changing it to "Starlark" fixes it.

allevato:
Puzzling. I'm not sure what would cause your workspace to have a different configuration for those particular files, unless you had another extension installed that was colliding with ours, or if there was some other old state interfering with it.

It would be interesting to know if this happens on a completely fresh VS Code + vscode-bazel install.

laurentlb:
I've disabled all other extensions. The extension was just installed:

```
Name: vscode-bazel
Id: bazelbuild.vscode-bazel
Description: Bazel BUILD integration
Version: 0.1.0
Publisher: The Bazel Team
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=bazelbuild.vscode-bazel
```
and VS Code looks up-to-date:
```
Version: 1.30.2
Commit: 61122f88f0bf01e2ac16bdb9e1bc4571755f5bd8
Date: 2019-01-07T22:49:48.319Z
Electron: 2.0.12
Chrome: 61.0.3163.100
Node.js: 8.9.3
V8: 6.1.534.41
OS: Linux x64 4.18.10-1rodete2-amd64
```

If you have any other idea, I can investigate tomorrow.

allevato:
@laurentlb, did you ever have any luck resolving this?

jxramos:
My experience is inverted. I'm on a macos.

I'm finding that *.bazel files are not syntax highlighted by default but BUILD files render fine.

[...]

PR #195, opened 2020-04-30
syntaxes: synchronize fileTypes with those indicated in package.json

slimsag:
This synchronizes the fileTypes in the syntax (which is more correct and helps those trying to use the syntax outside of VS Code) with those indicated in package.json

Note: In the case of `BUILD`, for example, which is not an extension, modern interpreters for tmLanguage support non-extension filenames here per https://forum.sublimetext.com/t/tmlanguage-and-filename-patterns/383

### DIFF
--- a/syntaxes/starlark.tmLanguage.json
+++ b/syntaxes/starlark.tmLanguage.json
@@ -9,7 +9,7 @@
         "star",
         "BUILD.bazel",
         "WORKSPACE",
-        "WORKSPACE.bazel",
+        "WORKSPACE.bazel"
     ],
     "patterns": [
         {
--- a/syntaxes/starlark.tmLanguage.json
+++ b/syntaxes/starlark.tmLanguage.json
@@ -2,10 +2,15 @@
     "name": "Starlark",
     "scopeName": "source.starlark",
     "fileTypes": [
+        "BUILD",
+        "WORKSPACE",
         "bzl",
+        "sky",
         "star",
         "build_defs",
-        "BUILD"
+        "BUILD.bazel",
+        "WORKSPACE",
+        "WORKSPACE.bazel",
     ],
     "patterns": [
         {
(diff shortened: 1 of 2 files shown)


syntaxes/starlark.tmLanguage.json
```diff
@@ -2,7 +2,10 @@
     "name": "Starlark",
     "scopeName": "source.starlark",
     "fileTypes": [
-        "bzl"
+        "bzl",
+        "star",
+        "build_defs",
```
laurentlb, line 7: I don't think `build_defs` is used.

slimsag: In specific, https://please.build/ declares build_defs files in Starlark/bazel syntax

slimsag:
@allevato that makes sense, and thanks for pointing me to that package.json!

I'm actually consuming `syntaxes/starlark.tmLanguage.json` outside of VS Code, since it appears to be the most well-maintained tmLanguage definition for Bazel's Starlark syntax.

In specific, I am consuming it in [syntect_server](https://github.com/sourcegraph/syntect_server) which is what we use at Sourcegraph for syntax highlighting and `fileTypes` in the syntax is to associate with specific files and also what is used if anyone in other editors (e.g. Sublime, etc) would discover similarly. So this change is actually to [improve Bazel highlighting in Sourcegraph](https://github.com/sourcegraph/sourcegraph/issues/8123) in specific.

I'll update package.json and starlark.tmLanguage.json to reflect the same set of filenames and extensions, which is definitely more correct -- but I cannot comment on whether it solves #54 specifically

slimsag:
Updated, please take a look and let me know what you think!

slimsag:
Makes sense, I've gone ahead and removed build_defs -- PTAL

slimsag:
I've also updated the PR description and title to reflect what this change actually does now.

allevato:
@googlebot rescan (just to make the bot status go green)

PR #211, opened 2020-08-16
Support ${workspaceFolder} for buildifier path

dapirian:
https://github.com/bazelbuild/vscode-bazel/issues/180

### DIFF
--- a/src/buildifier/buildifier.ts
+++ b/src/buildifier/buildifier.ts
@@ -165,7 +165,10 @@ export function getDefaultBuildifierExecutablePath(): string {
   const bazelConfig = vscode.workspace.getConfiguration("bazel");
   const buildifierExecutable = bazelConfig.get<string>("buildifierExecutable");
   if (vscode.workspace.rootPath !== undefined) {
-    buildifierExecutable.replace("${workspaceFolder}", vscode.workspace.rootPath);
+    buildifierExecutable.replace(
+      "${workspaceFolder}",
+      vscode.workspace.rootPath
+    );
   }
   if (buildifierExecutable.length === 0) {
     return "buildifier";


src/buildifier/buildifier.ts
```diff
@@ -164,6 +164,12 @@ export function getDefaultBuildifierExecutablePath(): string {
   // just use "buildifier" as the default and get it from the system PATH.
   const bazelConfig = vscode.workspace.getConfiguration("bazel");
   const buildifierExecutable = bazelConfig.get<string>("buildifierExecutable");
+  if (vscode.workspace.rootPath !== undefined) {
+    buildifierExecutable.replace(
+      "${workspaceFolder}",
+      vscode.workspace.rootPath
```
thomasvl, line 170: Is the behavior of this consistent for a trailing (or lack of trailing) `/`'?

i.e. - should folks use `${workspaceFolder}/something` or `${workspaceFolder}something`?

Maybe tweak https://github.com/bazelbuild/vscode-bazel/blob/master/package.json#L99 to document this support?

allevato:
I worry about trying to support just one of the variables that are available for task substitution—it opens the door to "why isn't this one, or that one, supported?" Even `${workspaceFolder}` can take multiple forms; for example, if you have multiple workspace roots, you can distinguish them by writing `${workspaceFolder:Name}`.

There's also a related discussion in an older VS Code issue (https://github.com/microsoft/vscode/issues/57551) about why this substitution doesn't automatically occur in settings, but it's light on details; they just say it would open a door that they don't want to open 🤷 

But my understanding from reading that issue is that people should **not** be writing `${workspaceFolder}` into paths in their settings. Even if an extension happens to support it by accident by virtue of the fact that that string ends up **directly** in a path passed to the task resolver, it won't work if you tried to write a task that referenced that setting. For example, unless something has changed since that issue was closed, if someone wrote a task that used `${bazel.executable}` and that setting contained `${workspaceFolder}`, the workspace folder wouldn't get resolved.

So even though some other extensions like clang-format support this using a subset of substitutions, is this something we should be encouraging?

Separately, I think we need to clarify what the expected behavior is for multi-root workspaces as well.

src/buildifier/buildifier.ts
dapirian: Folks should use `${workspaceFolder}/something` - and this is consistent with other usage of `${workspaceFolder}` in bazel

dapirian:
> I worry about trying to support just one of the variables that are available for task substitution—it opens the door to "why isn't this one, or that one, supported?" Even `${workspaceFolder}` can take multiple forms; for example, if you have multiple workspace roots, you can distinguish them by writing `${workspaceFolder:Name}`.
> 
> There's also a related discussion in an older VS Code issue ([microsoft/vscode#57551](https://github.com/microsoft/vscode/issues/57551)) about why this substitution doesn't automatically occur in settings, but it's light on details; they just say it would open a door that they don't want to open 🤷
> 
> But my understanding from reading that issue is that people should **not** be writing `${workspaceFolder}` into paths in their settings. Even if an extension happens to support it by accident by virtue of the fact that that string ends up **directly** in a path passed to the task resolver, it won't work if you tried to write a task that referenced that setting. For example, unless something has changed since that issue was closed, if someone wrote a task that used `${bazel.executable}` and that setting contained `${workspaceFolder}`, the workspace folder wouldn't get resolved.
> 
> So even though some other extensions like clang-format support this using a subset of substitutions, is this something we should be encouraging?
> 
> Separately, I think we need to clarify what the expected behavior is for multi-root workspaces as well.

_Many_ other extensions support this. I'm shocked vscode doesn't handle it themselves, it's annoying that they don't. We have a hermetic bazel and buildifier in our repo and I want the workspace to point to the in-repo versions. Without this (or jut supporting a relative path from the vscode workspace root) I can't do that. I don't mind any solution to this, as long as it's some solution. Happy to close this in favor of supporting a vscode-workspace-relative-path (which by the way, the `bazel` path handles, but the `buildifier` path doesn't, I didn't dig in enough to see how that's done)

PR #228, opened 2020-12-30
Fix "no-floating-promises" violations

hensom:
This change adds support for the "no-floating-promises" lint check.
Adding this check required the commandline tslint to explicitly
specify the project, which is required to enable type checking.

Without this change, simply adding the lint check would result in the
following warning:

 - Warning: The 'no-floating-promises' rule requires type information.

I also fixed all existing violations in the codebase.


hensom:
@googlebot I signed it!

PR #209, opened 2020-08-12
Code completion

alexfrasson:

PR #246, opened 2021-09-22
Code completion (review comments addressed)

coeuvre:
Based on #209. Addressed the review comments.


coeuvre:
@alexfrasson I have forked your patch in #209, addressed the review comments and created this PR. If you are OK with this, please reply with `@googlebot I consent`. Thanks!

philwo:
I'm manually overriding the CLA check for this PR, because
- this PR consists of three unmodified commits by @alexfrasson from https://github.com/bazelbuild/vscode-bazel/pull/209/commits for which we already have a signed CLA
- the other two comments are from @coeuvre who is a Googler, so the CLA doesn't apply.

@coeuvre I think we are good to merge this.

PR #253, opened 2021-10-29
Install missing packages to satisfy build.sh

fweikert:
Progress towards https://github.com/bazelbuild/vscode-bazel/issues/252


jmhodges:
I fixed those permission errors in #258 with a small change to `scripts/build.sh`. If you include that, you should be able to update your PR successfully.

jfirebaugh:
This seems to be fixed. `vsce package` currently completes without error.

PR #256, opened 2022-01-01
upgrade path-parse explicitly

jmhodges:
Any other changes to packages.json adds a downgrade (for some reason) to
path-parse 1.0.6 to the packages-lock.json. That's not ideal and, on top
of that, 1.0.6 has a vulnerability in it.
https://github.com/advisories/GHSA-hj48-42vr-x3v9
    
To prevent that from happening, we upgrade to 1.0.7 explicitly in the
lockfile.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,12 @@
 {
     "name": "vscode-bazel",
-    "version": "0.4.1",
+    "version": "0.5.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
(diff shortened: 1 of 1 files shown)


jmhodges:
This downgrade seems to happen anytime you run `npm install` at all.

jmhodges:
Oh, I've also had to include #258 in order to get the CI past the pbjs and pbts permission problems.

jfirebaugh (approved):
Thanks!

PR #258, opened 2022-01-01
workaround pbjs and pbts not having exec perms

jmhodges:
pbjs and pbts don't have the executable bit set on them when installed
into node_modules (possibly only on macos?). So, folks who run `npm run
compile` will get hit by needing to `chmod +x` those files and then do
it again whenever vscode-bazel upgrades to a new version of those.

Instead, we just call node explicitly in build.sh


coeuvre (approved):

jfirebaugh:
Sorry for the delayed response here. In the meantime it looks like this was fixed (in a slightly different way) in ffd1450a4222825db11533aac5f5c7aa51ef50eb.

Issue #265, opened 2022-05-13
Command fails from extension but succeeds in my terminal


paulherman:
I seem to be getting errors when using almost any Bazel command from the palette, as shown below.

```
rejected promise not handled within 1 second: Error: Command failed: bazel --output_base=<redacted path> query kind('.* rule', ...) --output=proto
Starting local Bazel server and connecting to it...
Loading: 0 packages loaded
ERROR: error loading package 'bazel-plh/external/bazel_tools/tools/cpp': at D:/paul/plh/bazel-plh/external/bazel_tools/tools/cpp/compiler_flag.bzl:17:6: Label '//tools/cpp:toolchain_utils.bzl' is invalid because 'tools/cpp' is not a package; perhaps you meant to put the colon here: '//:tools/cpp/toolchain_utils.bzl'?
Loading: 68 packages loaded
    currently loading: bazel-plh/external/bazel_tools/tools/android ... (38 packages)
Loading: 68 packages loaded
    currently loading: bazel-plh/external/bazel_tools/tools/android ... (38 packages)
E @ console.ts:137
```

I've tried running the same command in the terminal and it succeeds if and only if I'm adding quotations around the query.

aschleck:
I think what's happening is you need to run a build or something to actually create all necessary analysis files.

Eg
```
bazel --output_base=/tmp/4d8027e6d32eeb7619bc14d5897e6dd1 query '//...:*' --output=package
```
will fail, but if you run
```
bazel --output_base=/tmp/4d8027e6d32eeb7619bc14d5897e6dd1 build //:some_target
```
and then run
```
bazel --output_base=/tmp/4d8027e6d32eeb7619bc14d5897e6dd1 query '//...:*' --output=package
```
again it will work. If you turn on "Queries Share Server" and force VS Code to build some target, I would expect this to work. It at least fixes "Bazel Build Targets" in the sidebar which has never worked for me.

cono:
Same issue here. Tried with "Queries Share Server" and without, same issue.
in my case it blows up with error:
```
[31m[1mERROR: [0merror loading package '<redacted>/external/io_bazel_rules_scala/<redacted>': Label '//scala:providers.bzl' is invalid because 'scala' is not a package; perhaps you meant to put the colon here: '//:scala/providers.bzl'?
```

Where scala dep is:
```
# Scala Rules
git_repository(
    name = "io_bazel_rules_scala",
[...]

PR #270, opened 2022-07-25
Make quereis to share the same server by default

limdor:
The bazel query used to introspect the targets available assumes that the generated files are
already there.
This creates the problem that you have to first generate the files once.
https://github.com/bazelbuild/vscode-bazel/issues/265

Not sharing the same server makes it harder because the output is a random one.
This means that for a normal user, it will be hard to make it work when query does not
use the same server like your normal bazel queries.
Because of that, it is better to have the queriesShareServer to true by default
and the expert users can set it to false if they want.


package.json
```diff
@@ -107,7 +107,7 @@
                 },
                 "bazel.queriesShareServer": {
                     "type": "boolean",
-                    "default": false,
+                    "default": true,
                     "description": "Use the same Bazel server for queries and builds. By default, vscode-bazel uses a separate server for queries so that they can be executed in parallel with builds. You can enable this setting if running multiple Bazel servers has a negative performance impact on your system, but you may experience degraded performance in Visual Studio Code for operations that require queries."
```
limdor, line 111: Just one note, the description would have to be adapted because it assumes that default is false. Before I do that I would like to know the opinion of the contributors if they would agree on having default to true or not.

limdor: @coeuvre @philwo any feedback?

cpsauer:
+1 on this. Otherwise a bunch of features are broken by default as in @265. Any chance we could get a review?

Thanks,
Chris
(ex-Googler)

limdor:
@jfirebaugh @coeuvre could some of you comment on this PR? Or should I just close it?

jfirebaugh:
I'm up for changing this default and seeing how it works. I'm going to put out a 0.8.1 release with a fix for a regression in 0.8.0, and then I'll merge this.

package.json
```diff
@@ -154,7 +154,7 @@
...
```
limdor, line 158:
```suggestion
                    "description": "Use the same Bazel server for queries and builds. By default, vscode-bazel uses the same server for queries and builds in order to have it easier for the user to set up the plugin. However, you may experience degraded performance in Visual Studio Code for operations that require queries. You can disable this setting so that queries and builds can be executed in parallel. Keep in mind that you might get some errors after disabling this setting if you do not configure it properly. See https://github.com/bazelbuild/vscode-bazel/issues/265."
```

cpsauer:
Still down to merge @jfirebaugh--or is this waiting on @limdor?

limdor:
If that is the case let me know because from my side everything is done. In the moment I das informed there is interest to merge it I also update the message to fit to the new default.

cameron-martin:
> This creates the problem that you have to first generate the files once.

I think this diagnosis is wrong. The issue in #265 looks to be caused by bazel traversing the convenience symlinks when gathering packages (see https://github.com/bazelbuild/bazel/issues/10653). I've opened up [a PR in Bazel](https://github.com/bazelbuild/bazel/pull/21005) to mitigate this but I think disabling this setting by default is a good option in the meantime.

cpsauer:
Yay! Thanks, Cameron. I think that'll help a lot of new folks get benefit out of the box--once released.

PR #282, opened 2022-10-09
Expand matching files to *.bazel

dierksen:
With the addition of `MODULE.bazel` to the mix, it seems to make sense to match all files with this extension.


jfirebaugh (approved):
Makes sense to me.

PR #299, opened 2023-03-27
Add a new command to run targets without arguments

romannikov:
This implementation doesn't support to pass arguments for the `blaze run //path:taget` command, but it can be implemented in the following changes.


jfirebaugh (approved):
Looks good to me, thanks!

PR #305, opened 2023-11-05
Fix candidate target name parsing for completion

### LINKED ISSUE #304: Completion doesn't work inside `$(location)`, `$(execpath)`, etc.
Completion doesn't seem to work when the target name is not quoted.

kon72:
This change fixes a bug where the target completion doesn't work when the target name is not quoted (like inside `$(location)`).

Fixes #304


jfirebaugh (approved):
Makes sense, thanks!

PR #306, opened 2023-11-05
Add support for relative target completion

kon72:
This change enables the target completion to work with relative label starting with colon.

Depends on #305


jfirebaugh:
Thanks! I took the liberty of rebasing this on current master after merging #305. Do you think this and #307 are ready for review?

kon72:
HI @jfirebaugh,

Thanks for the rebasing! I think this change is good to go.

For #307, I'm wondering if it may introduce some performance degradation, so I will do some more investigation and add an option for users to opt out of external rule completion.

jfirebaugh (approved):
Thank you! Out of curiosity, are you familiar with testing VS Code extensions? (I've never done it before.) Code completion seems like the sort of thing that would be nice to have tests around. No worries if not!

cpsauer:
@kon72, thanks for working on these!

PR #313, opened 2023-12-08
Identify WORKSPACE.bzlmod as Starlark

lalten:
See https://bazel.build/external/migration#:~:text=in%20WORKSPACE%2C-,WORKSPACE.bzlmod,-and%20each%20Bazel


jfirebaugh (approved):
Thanks!

PR #317, opened 2023-12-19
Add support for launching an LSP

### LINKED ISSUE #1: Implement language server for Starlark
First off, I'm really excited to see the Bazel team creating a repository for vscode support and I would love to contribute to it!

Secondly, Is this Repository going to be developing a [Language Server Protocol](https://langserver.org/) implementation? Seems like it could be a great opportunity to develop a resource that could be reused in other editors as well.

cameron-martin:
This adds an option for launching [facebook's starlark LSP server](https://github.com/facebookexperimental/starlark-rust). Currently this has to be installed separately and is very experimental since the LSP server seems more of a proof-of-concept. This disables some functionality of the extension (go to definition and the symbol provider), since it is provided by the LSP instead.

Resolves #1


package.json
```diff
@@ -198,6 +203,27 @@
                     "type": "string",
                     "default": "...:*",
                     "description": "A [query language expression](https://bazel.build/query/language) which determines the packages displayed in the workspace tree and quick picker. The default inspects the entire workspace, but you could narrow it. For example: `//part/you/want/...:*`"
+                },
+                "bazel.lsp.enabled": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enable the LSP (experimental)"
```
daivinhtran, line 210: Could we add a note here to remind that `lsp.command` needs to be set in order for it to work?

We can also imply it to be true when `lsp.command` is set.

README.md
```diff
@@ -49,6 +49,13 @@ Clicking the "Stop" button in the debugger will kill the Bazel process being
 debugger, allowing you to halt the current build. The Bazel server, however,
 will continue running.
 
+## Using the LSP (Experimental)
+
+This extension can use [Facebook's starlark project](https://github.com/facebookexperimental/starlark-rust) as a language server.
+
+1. Install the LSP using cargo: `cargo install starlark_bin`
```
daivinhtran, line 56: We also need to add the binary path to `PATH`.

package.json
cameron-martin: `lsp.command` defaults to the executable name that cargo will install it under (`starlark`), so setting this shouldn't be necessary unless you're doing something strange.

README.md
cameron-martin: `cargo install` will put this in `~/.cargo/bin`, and this should be added to your path during installation of cargo.

package.json
daivinhtran: I see. I also meant to ask that should we remove the default `starlark` from `lsp.command`?

Instead of having these scenarios:

1.
```
enabled: False
command: "starlark" // no-op
```
2.
```
enabled: True
command: "starlark"
```
, we'll just have
1.
```
command: undefined // disabled
```
2.
```
command: "starlark"
```

Just one less configuration to simply it a little bit in case users don't see the README before seeing configurations. By forcing users to specify `starlark` explicitly, they're more aware that they need to build/install the `starlark` binary in the first place. If not, `starlark` stays undefined until `cargo` builds it.

daivinhtran:
Btw, CI fails because of some linting errors

```

/workdir/src/extension/extension.ts:18:3
--
  | ERROR: 18:3   ordered-imports           Named imports must be alphabetized.
  | ERROR: 187:5  object-literal-sort-keys  The key 'args' is not sorted alphabetically
```

I ran into similar issue earlier. Fixing them might fix CI.

package.json
cameron-martin: Personally I think having a single config option mean two different things is a bit confusing, and checking a checkbox is easier than writing "starlark". I think the case where the user hasn't installed the binary should be handled with an informative error message, but this is pretty experimental so I'm hesitant to add that sort of polish yet.

daivinhtran: > Personally I think having a single config option mean two different things is a bit confusing

It doesn't necessarily mean two different things. It can be interpreted as

```
If lsp command is specified, use it to start a lsp client.
Otherwise, don't
```

This is so that we don't need to add the informative error message. The config is then self-explanatory.

Anyways, this is just my opinion. I checked out your branch and it worked great! Thanks!

daivinhtran:
FYI I tried enabling LSP again today to work in the https://github.com/bazelbuild/rules_rust repo. I noticed the "BAZEL BUILD TARGETS" window didn't load. I then disabled LSP and the "BAZEL BUILD TARGETS" loaded correctly.

Another bug I noticed is that repo like https://github.com/bazelbuild/examples will crash with LSP being enabled because the root directory doesn't have a WORKSPACE and/or BUILD file.

daivinhtran:
@coeuvre @jfirebaugh any objection on merging this experimental feature?

coeuvre:
SGTM to merge this feature. @jfirebaugh what do you think?

jfirebaugh (approved):
Yeah, let's do it.

PR #322, opened 2023-12-20
Prevent automatic MODULE.bazel file creation

daivinhtran:
Bazel 7 [enables bzlmod](https://bazel.build/about/roadmap#bzlmod_external_dependency_management_system) by default which results to the automatic generation of `MODULE.bazel` files when the extension runs bazel command in the background.

From a VSCode extension standpoint, it's a bad UX (IMO) to create new files unintentionally. If Bazel wants to enforce having `MODULE.bazel` files moving forward, that is already covered when users run Bazel commands locally.

In the future, we can gate this flag behind a boolean config when bzlmod is fully supported https://github.com/bazelbuild/vscode-bazel/issues/294.

### DIFF
--- a/src/bazel/bazel_command.ts
+++ b/src/bazel/bazel_command.ts
@@ -75,10 +75,21 @@ export abstract class BazelCommand {
       vscode.workspace.getConfiguration("bazel.commandLine");
     const startupOptions = bazelConfigCmdLine.get<string[]>("startupOptions");
 
+    // Options that are applied to all Bazel commands
+    const defaultOptions = [
+      // See https://bazel.build/versions/6.4.0/external/lockfile
+      // Prevent the extension from automatically creating the MODULE.bazel file
+      // on bazel v7 or newer which enable bzlmod by default
+      // Bazel can still enforce creating the MODULE.bazel automatically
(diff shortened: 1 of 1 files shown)


coeuvre (approved):

jfirebaugh (commented):
Hmm, I'm not entirely convinced this is the right thing to do. Won't it mean that any Bazel commands executed by the extension do not benefit from the advantages of lockfiles (reproducible builds, efficient resolution skipping, stability and risk reduction)?

daivinhtran:
@jfirebaugh I'm not disagreeing. However, MODULE.bazel (and bzlmod) don't seem to be handled that well by the extension per https://github.com/bazelbuild/vscode-bazel/issues/294.

However, I'm ok with reverting the PR. Your suggestion sounds like a better practice to me.

cameron-martin:
The unwanted creation of MODULE.bazel files only happens to me when the repo contains other external repositories that aren't meant to be used as root workspaces. However, I think this is due to the cwd that bazel uses to run commands.

See https://bazelbuild.slack.com/archives/C014RARENH0/p1702897315236089 for more info.

PR #333, opened 2024-01-22
Add @cameron-martin to CODEOWNERS

jfirebaugh:
Per https://github.com/bazelbuild/vscode-bazel/issues/241#issuecomment-1900405902. Thank you for offering!

I think there's a second step that I don't have permission to do, which is to add @cameron-martin as a collaborator with write access. At least that's what GitHub is telling me:

Screenshot 2024-01-22 at 2 56 44 PM

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @coeuvre @jfirebaugh
+* @coeuvre @jfirebaugh @cameron-martin


meteorcloudy:
I sent an invitation to @cameron-martin

meteorcloudy (approved):

cameron-martin:
@coeuvre I think this needs a review from you too, since you're the only other code owner.

coeuvre (approved):

PR #339, opened 2024-02-03
Increase vscode engine version

cameron-martin:
When running `vsce package`, it complains that the versions of `engines.vscode` and `@types/vscode` are incompatible. This updates `engines.vscode` to match the version of `@types/vscode`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     },
     "icon": "media/bazel-logo.png",
     "engines": {
-        "vscode": "^1.30.0"
+        "vscode": "^1.85.0"
     },
     "categories": [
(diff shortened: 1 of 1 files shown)


cameron-martin:
What's the support meant to be on vscode versions? Should we actually downgrade `@types/vscode` to a newer version?

This mismatch was caused by upgrading it in #317. I needed to do this to use the newest version of `vscode-languageclient`, but possibly I didn't need to upgrade it to latest.

jfirebaugh (approved):
Given that vscode has evergreen auto-updating behavior, I don't see an issue with keeping the engine version relatively recent.

PR #345, opened 2024-02-27
Enable the "Restart LSP" command only if a LSP is configured

vogelsgesang:
This commit hides the "Restart LSP" command from the Command Palette if the LSP is disabled. Otherwise, trying to use the "Restart LSP" command without any LSP being configured leads to a user-visible error.

As part of this, I also renamed `vscodeBazelHaveBazelWorkspace` to `bazel.haveWorkspace` such that the context variables all share a common `bazel.*` prefix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,8 @@
             {
                 "category": "Bazel",
                 "command": "bazel.lsp.restart",
-                "title": "Restart language server"
+                "title": "Restart language server",
+                "when": "bazel.lsp.enabled"
             }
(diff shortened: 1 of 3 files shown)


cameron-martin (approved):

vogelsgesang:
Thanks for your review, @cameron-martin !

It seems I don't have the necessary permissions to merge this PR myself. What is the usual workflow in this repository? Can you merge this on my behalf?

PR #346, opened 2024-02-27
Support `bazel` tasks from `tasks.json`

vogelsgesang:
User-facing changes
-------------------

This commit fixes two user-facing bugs:
* The "Run Task" command provides a list of recently used tasks. This list also contains the build/test tasks started from the "Bazel Build Targets" tree view. However, trying to rerun any of those tasks resulted in an error message.
* The extension already contributed a `taskDefinition` for bazel tasks. However, trying to actually define such a task in the `tasks.json` resulted in the same error message.

In both cases, the user was presented with the error message

> There is no task provider registered for tasks of type "bazel"

This commit fixes the issue, such that rerunning tasks from the history and configuring tasks through `tasks.json` now works properly.

Furthermore, the `taskDefinition` inside the `package.json` was improved:
* It now lists the available `command` options, such that we have auto-completion within the `tasks.json` editor.
* It properly constraints the `targets` entries to be string-typed.
* It exposes the `options` argument which allows to specify additional command line parameters.

Code changes
------------

Fixing the issue mainly required registering a `TaskProvider` for `bazel` tasks. The `TaskProvider` currently does not auto-detect the available tasks, but only implements `resolveTask`.

The notifications in `onTaskProcessEnd` had to be refactored: So far, the `BazelTaskInfo` was attached right during task creation. However, this didn't work for the new `resolveTask`. `resolveTask` apparently isn't allowed to change a task's definition. To make notifications work, the `BazelTaskInfo` is now attached in the `onTaskStart` callback.

Many of the fields of the `BazelTaskInfo` were unused. This commit removes `commandOptions`, `processId`, `exitCode` all of which were stored but never read. Doing so, it turned out that the `onTaskProcessStart` was no longer necessary.

Lastly, the callbacks used for task notifications were moved from `extension.ts` to `bazel_task_info.ts` and activation of all task-related functionality was centralized in the `activateTaskProvider` function

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ This extension provides support for Bazel in Visual Studio.
   clicking on the targets
 - **Buildifier** integration to lint and format your Bazel files (requires that
   [Buildifier](https://github.com/bazelbuild/buildtools/releases) be installed)
+- **Bazel Task** definitions for `tasks.json`
 - Debug Starlark code in your `.bzl` files during a build (set breakpoints, step
   through code, inspect variables, etc.)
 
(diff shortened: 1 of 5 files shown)


jfirebaugh (approved):
Thanks!

PR #350, opened 2024-02-29
Support running buildifier through Bazel

### LINKED ISSUE #185: Support using @com_github_bazelbuild_buildtools//buildifier instead of download buildifier manully
Buildifier can be run directly with bazel:

```
bazel run @com_github_bazelbuild_buildtools//buildifier -- -version
```

is it possible to use @com_github_bazelbuild_buildtools//buildifier if it was found in WORKSPACE?

### LINKED ISSUE #329: support relative paths for `bazel.buildifierExecutable`
Currently buildifier is executed from the disk root, which means that if you configure a path like `"bazel.buildifierExecutable": "./path/to/buildifier"` then the extension will execute `/path/to/buildifier` - which likely doesn't exist.

It would be good if the CWD for the execution could be set to the workspace root to enable relative path config. This is similar to #132 but doesn't require any variable resolution.

vogelsgesang:
Buildifier is frequently installed through Bazel as described on https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md#setup-and-usage-via-bazel such that it can be run through `bazel run`.

This commit allows the `bazel.buildifierExecutable` setting to refer to a Bazel target. All paths starting with `@` are interpreted as Bazel target names and are executed through `bazel run`. Bazel targets could also start with `//` but we interpret those as normal file system paths. If someone wants to run a target in the own workspace, they can simply use, e.g., `@//:buildifier` instead of `//:buildifier`.

As a drive-by fix, I also fixed #329 since it was a one-line fix.

Fixes #185, #329

BEGIN_COMMIT_OVERRIDE
feat: Support running buildifier through Bazel (#350)

feat:  Support relative paths for bazel.buildifierExecutable (#350) 
feat: Pick up `.buildifier.json` configuration from the Bazel workspace root (#350)
END_COMMIT_OVERRIDE

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
                 "bazel.buildifierExecutable": {
                     "type": "string",
                     "default": "",
-                    "markdownDescription": "The name of the Buildifier executable. This may be an absolute path, or a simple name that will be searched for on the system path. If empty, \"buildifier\" on the system path will be used.\n\nBuildifier can be downloaded from https://github.com/bazelbuild/buildtools/releases.",
+                    "markdownDescription": "The name of the Buildifier executable. This may be an absolute path, or a simple name that will be searched for on the system path. Paths starting with `@` are interpreted as Bazel targets and are run via `bazel run`. If empty, \"buildifier\" on the system path will be used.\n\nBuildifier can be downloaded from https://github.com/bazelbuild/buildtools/releases.",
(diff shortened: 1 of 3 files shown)


jfirebaugh (approved):
Thank you!

Zemnmez:
Thank you for this!

PR #352, opened 2024-02-29
Enable incremental TypeScript builds

vogelsgesang:
Cuts the time of `npm run compile` from 9 seconds to 3 seconds

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,8 @@
         "sourceMap": true,
         "rootDir": ".",
         "alwaysStrict": true,
-        "allowJs": true
+        "allowJs": true,
+        "incremental": true,
     },
     "exclude": [
         "node_modules",


jfirebaugh (commented):
I'm mildly opposed to this. Every time I've tried to use incremental mode, I've ended up needing to turn it off because of weird bugs that lead to miscompilation or hard-to-debug behavior. It just doesn't seem to be a reliable feature. Plus it's not supported by rules_js, which we should be migrating to anyway.

cameron-martin:
> I'm mildly opposed to this. Every time I've tried to use incremental mode, I've ended up needing to turn it off because of weird bugs that lead to miscompilation or hard-to-debug behavior. It just doesn't seem to be a reliable feature. Plus it's not supported by rules_js, which we should be migrating to anyway.

If there's known correctness issues then I'd also be in favour of not enabling this for the sake of 6 seconds.

vogelsgesang:
Interesting... I never saw incremental build issues on any of my other Typescript projects. But given this does not align with the goals around `rules_js`, I agree we shouldn't merge it

On a related note: Are there reasons why the `strict` setting is not enabled in the `tsconfig.json`? I noticed a lot of issues diagnosed in particular by `strictNullChecks`. Would you welcome a change which enables `strict` type checking?

jfirebaugh:
I would love a change to enable `strict`!

PR #354, opened 2024-03-03
feat: Introduce `pickTarget` and `pickPackage` command variables

vogelsgesang:
This commit introduces the `bazel.pickTarget` and `bazel.pickPackage` command variables. Those commands can be used from `tasks.json` and `launch.json` to prompt the user to select a Bazel target or package. The available choices are determined by a user-specified Bazel query.

The implementation reuses the existing quick-pick functionality from `bazel_quickpick.ts`. The `wrapQuickPick` function parses the arguments passed from the `task.json` and then forwards to the existing quickpicks.

Those utility functions were restructured as part of this commit:
1. The internal helpers `queryWorkspaceQuickPickPackages` and `queryWorkspaceQuickPickTargets` were inlined into `queryQuickPick{Packages,Targets}` as the internal helper functions got in the way of other changes, but didn't deduplicate any code as they only had one caller.
2. The `queryQuickPick{Packages,Targets}` take an object with named parameters now.
3. `queryQuickPickPackage` now also accepts a `query` parameter instead of using the `queryExpression`. Both `queryQuickPickPackages` and `queryQuickPickTargets` default to the query `//...` if no query string is provided by the caller.

Point (3) also is a user-facing change. Commands like "Build package" and "Test package" now use `//...` instead of the configured `queryExpression`. This is in sync with other commands like "Test target" which uses the hardcoded query `kind('.*_test rule', ...)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,19 @@ Bazel tasks can be configured from the `launch.json` using the following structu
       "label": "Check for flakyness",
       "type": "bazel",
       "command": "test",
-      "targets": ["//my/package:integration_test"],
+      "targets": ["${input:pickFlakyTest}"],
       "options": ["--runs_per_test=9"]
+    }
+  ],
+  "inputs": [
+    {
+      "id": "pickFlakyTest",
+      "type": "command",
+      "command": "bazel.pickTarget",
+      "args": {
+        "query": "kind('.*_test', //...:*)",
+        "placeHolder": "Which test to check for flakyness?"
+      }
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ This extension can use [Facebook's starlark project](https://github.com/facebook
 
 ## Bazel tasks
 
-Bazel tasks can be configured from the `launch.json` using the following structure:
+Bazel tasks can be configured from the `tasks.json` using the following structure:
 
 ```json
 {
--- a/src/assert.ts
+++ b/src/assert.ts
@@ -22,5 +22,6 @@ export function assert(value: boolean): asserts value {
           "https://github.com/bazelbuild/vscode-bazel/issues",
       );
     }
+    throw new Error("Assertion violated.");
   }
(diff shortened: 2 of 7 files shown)


vogelsgesang:
@cameron-martin @jfirebaugh do you have time to review this? In case you currently don't have time to review in detail, I would appreciate high-level feedback whether you consider this functionality as valuable in general 🙂

cameron-martin:
I've been on holiday but I reckon I'll be able to take a look next week. The functionality does look useful though.

vogelsgesang:
By the way: I think long-term it would make sense to also add support for `cquery` and parameters such as [--experimental_explicit_aspects](https://bazel.build/reference/command-line-reference#flag--experimental_explicit_aspects), [--universe_scope](https://bazel.build/reference/command-line-reference#flag--universe_scope), [--infer_universe_scope](https://bazel.build/reference/command-line-reference#flag--infer_universe_scope), [--noimplicit_deps](https://bazel.build/reference/command-line-reference#flag--implicit_deps). But I will leave that for a different commit.

README.md
cameron-martin, line 1: I know this is existing, but this JSON should go in the `tasks.json` file, not `launch.json` as mentioned here.

cameron-martin (changes requested):
Apart from the two other comments, looks good.

src/assert.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from "vscode";
+
+let assertionFailureReported = false;
+
+/**
+ * Asserts that the given value is true.
+ */
+export function assert(value: boolean): asserts value {
```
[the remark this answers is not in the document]
vogelsgesang: good catch! Fixed

README.md
vogelsgesang: I actually introduced this issue myself in a previous commit... more than happy to fix it :)

cameron-martin (approved):

Issue #216, opened 2020-09-08
Warnings from finding build targets


efagerho:
I just tried out Bazel for the first time for a Java project and I ran into an issue that I've never seen with C++ and Python projects. Basically, when I try to hit the "refresh" button on the "Bazel build targets" side panel, I get an error popup like the following:

`Command failed: bazel --output_base=/var/folders/cj/1fv063dx1772zbpnfpzvj_c00000gn/T/d855bbcad42db418fe212896af750013 query ...:* --output=package Loading: 0 packages loaded ERROR: error loading package 'bazel-bazeljava/external/rules_jvm_external': Every .bzl file must have a corresponding package, but '//:private/versions.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist. Loading: 12 packages loaded currently loading: bazel-bazeljava/external/local_config_cc ... (9 packages) Loading: 12 packages loaded currently loading: bazel-bazeljava/external/local_config_cc ... (9 packages)`

The error is slightly different each time and seems to be located in different external packages. Maybe my version of Bazel has some issues with the external dependencies that are being loaded.

Bazel version: 3.5.0
VSCode version: 1.48.2

Steps to reproduce:

1. Clone the repository: https://github.com/efagerho/bazeljava
2. Run: bazel build ...
3. Open repository in VSCode
4. Click the refresh button under "Bazel build targets"

I don't see anything wrong with my project WORKSPACE file. If I manually run `bazel query ...:*` in the root, I don't see any errors and all build targets are properly listed. Seems like correct behavior would be for VSCode to list all the local targets in my workspace and not fail because of issues with external dependencies.

ian-h-chamberlain:
I was able to fix this by deleting the folder listed as the `--output_base=` argument, which forced the extension to re-fetch the apparently missing dependencies.
Presumably running the "Bazel: Clean" command would have the same effect. Hopefully this helps anyone else running into this issue! 

My extension + Bazel info:
```
$ bazel version 
Bazelisk version: 1.6.1
Build label: 3.3.0
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Jun 17 12:59:56 2020 (1592398796)
Build timestamp: 1592398796
Build timestamp as int: 1592398796
```
Extension
```
Name: Bazel
Id: bazelbuild.vscode-bazel
[...]

PR #355, opened 2024-03-12
Revert "Make queries share the same server by default (#270)"

cameron-martin:
This reverts commit 64c9d7eb41bed5a3b9ab67a6423f1f7eee56ed65.

The underlying issue that this addresses is now fixed in Bazel 7.1.0. See #216 and https://github.com/bazelbuild/bazel/pull/21505.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,8 +128,8 @@
                 },
                 "bazel.queriesShareServer": {
                     "type": "boolean",
-                    "default": true,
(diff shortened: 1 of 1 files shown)


jfirebaugh (approved):

cameron-martin:
Good point. I think displaying a message for older versions would be good, since this also has a manual workaround of adding the convenience symlinks to `.bazelignore`.

cameron-martin:
@vogelsgesang I've put up a PR to revert this change, plus add some docs in #367.

PR #357, opened 2024-03-17
Add `buildifier` Json File Configuration

CauhxMilloy:
This PR adds the `buildifierConfigJsonPath` setting, which allows a custom path to be set to pull the json configuration file for `buildifier` (this is often `.buildifier.json`). If the `buildifierConfigJsonPath` path is set, when `buildifier` is run within this extension, the `--config <path>` CLI arguments will be passed to `buildifier`.

While it could be said that this PR is now adding support for the `.buildifier.json` file -- it's only adding support for the customization of that file path. https://github.com/bazelbuild/vscode-bazel/pull/350 actually pulled out yet another drive-by fix (of https://github.com/bazelbuild/vscode-bazel/issues/208) by adding the working directory. In that PR, `cwd: vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath` is setting the working directory for the `buildifier` execution to be the local workspace. This means any local `.buildifier.json` file will automatically get pulled in and used. This PR is allowing further customization on top of that.

The logic of this PR was tested by manually editing the js of my local extension installation with the equivalent changes (that's how I found that `cwd: ...` ended up working with a local `.buildifier.json`).

### DIFF
--- a/src/buildifier/buildifier.ts
+++ b/src/buildifier/buildifier.ts
@@ -175,6 +175,18 @@ export function getDefaultBuildifierExecutablePath(): string {
 }
 
 /**
+ * Gets the path to the buildifier json configuration file specified by the
+ * workspace configuration, if present.
+ *
+ * @returns The path to the buildifier json configuration file specified in the
+ * workspace configuration, or an empty string if not present.
+ */
+export function getDefaultBuildifierJsonConfigPath(): string {
+  const bazelConfig = vscode.workspace.getConfiguration("bazel");
+  return bazelConfig.get<string>("buildifierConfigJsonPath");
+}
+
+/**
  * Executes buildifier with the given file content and arguments.
  *
  * @param fileContent The BUILD or .bzl file content to process, which is sent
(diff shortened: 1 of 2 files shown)


src/buildifier/buildifier.ts
```diff
@@ -191,6 +203,10 @@ export function executeBuildifier(
   return new Promise((resolve, reject) => {
     // Determine the executable
     let executable = getDefaultBuildifierExecutablePath();
+    const buildifierConfigJsonPath = getDefaultBuildifierJsonConfigPath();
+    if (buildifierConfigJsonPath.length !== 0) {
```
vogelsgesang, line 207: afaict, `bazelConfig.get<string>` and thereby `getDefaultBuildifierJsonConfigPath` can return `undefined`. Accessing `undefined.length` would lead to an error. We should probably use

```
  return bazelConfig.get<string>("buildifierConfigJsonPath") ?? "";
```

or something similar

CauhxMilloy: Looking at the [documentation](https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration), it does seem that `get<T>()` can return `undefined`. But why is this not an issue for the (existing) `getDefaultBuildifierExecutablePath()`? In any case, I'll go ahead and change this to use the overload with a default to fallback to `""` instead of `undefined`.

Issue #362, opened 2024-04-02
Integrate `bazel coverage` support


vogelsgesang:
The upcoming VS Code release will add support for code coverage (See https://github.com/microsoft/vscode/issues/123713). It probably makes sense to integrate this with `bazel coverage`, such that the code coverage is directly shown in the IDE

PR #366, opened 2024-04-05
feat: Basic test coverage support

vogelsgesang:
This commit lays the foundation for displaying coverage results from `bazel coverage`.

For the time being, the functionality is only exposed through the user-defined tasks in the `tasks.json`. It is thereby a bit hard to discover. But this is fine for the time being, because coverage still has a couple of rough edges anyway. As soon as it is more stable, we should add builtin commands and expose coverage runs also in the "Bazel Build Target" tree.

Changes in this commit:
* Bumps the VS Code version to 1.88, i.e. the first VS Code version which supports the test coverage API.
* Upgrades to ES2022.  I wanted to use `replaceAll` which was introduced in ES2021. VS Code 1.88 is based on Node 18 which in turn is based on V8 10.1. V8 10.18 supports ECMA-262 also known as ES2023.  However, ES2023 is not yet available a target language in the `tsconfig.json`.  Furthermore, Firefox does not fully support ES2023, yet. While web browsers are currently not relevant, they might become so in the future if we want to turn this into a browser-enabled VSCode extension. An upgrade to ES2021 would have been sufficient, but I went directly to ES2022 because it might some of the other new features might also turn out useful.
* Introduces a custom LCOV parser. I could not find any other high-quality open-source parser. E.g., most other parser don't properly parse function names with `:` and / or `,` in them.
* Introduces test cases for that custom LCOV parser.

Future work:
* Support for branch coverage
* Demangling of function names
* Builtin commands to trigger coverage runs & offer them in the "Bazel Build Tree"

Tested with: Java, C++, Go, Rust
Untested: Python, Swift, Kotlin, Scala and many more

This is the first step towards #362


vogelsgesang:
See https://github.com/microsoft/vscode/issues/209697 for a couple of screenshots

cameron-martin:
I've reviewed everything but the lcov parser, and it looks good. I've run out of time this evening, so I'll have to review the lcov parser another day.

vogelsgesang:
I am still interested in merging this, but I will wait with resolving the merge conflicts until after the review 🙂

cameron-martin (approved):
Sorry for the delay, looks good.

vogelsgesang:
No worries, I know this was a larger, more complicated commit. Thanks for taking the time to review this in-depth instead of just give me a "probably good-enough approval"! Good reviews are important to the project in the long run 🙂

PR #369, opened 2024-04-06
docs: Update options for language servers

cameron-martin:
Previously the docs suggested using facebook's proof-of-concept bazel language server, but there is a more up-to-date version available based on this. There is also another implementation, starpls, that is based on rust-analyzer. This also updates the default values of the lsp settings to not favour a particular language server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,12 +50,14 @@ Clicking the "Stop" button in the debugger will kill the Bazel process being
 debugger, allowing you to halt the current build. The Bazel server, however,
 will continue running.
 
-## Using the LSP (Experimental)
+## Using a language server (Experimental)
 
-This extension can use [Facebook's starlark project](https://github.com/facebookexperimental/starlark-rust) as a language server.
+Some of the functionality of this extension, such as go to definition and completions, can be replaced with a language server. There are currently two in-progress implementations of this:
 
-1. Install the LSP using cargo: `cargo install starlark_bin`
-2. Enable the LSP extension by setting `bazel.lsp.enabled` to `true`.
+- [bazel-lsp](https://github.com/cameron-martin/bazel-lsp): Based on Facebook's starlark project.
+- [starpls](https://github.com/withered-magic/starpls): Based on rust-analyzer.
+
+See the README of the corresponding repo for setup instructions.
 
 ## Bazel tasks
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Clicking the "Stop" button in the debugger will kill the Bazel process being
 debugger, allowing you to halt the current build. The Bazel server, however,
 will continue running.
 
-## Using a language server (Experimental)
+## Using a language server (experimental)
 
 This extension can use a language server for various features, such as go to definition and completions. There are currently two compatible language servers:
 
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ will continue running.
 
 This extension can use a language server for various features, such as go to definition and completions. There are currently two compatible language servers:
 
-- [bazel-lsp](https://github.com/cameron-martin/bazel-lsp): is based on Facebook's Starlark language server and extends it with additional, Bazel-specific functionality.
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -188,18 +188,15 @@
                 },
                 "bazel.lsp.command": {
                     "type": "string",
-                    "default": "starlark",
+                    "default": "",
```
vogelsgesang, line 191: One thing I was wondering: Should we maybe also remove the `bazel.lsp.enabled` setting? And instead just enable the LSP as soon as `bazel.lsp.command` is set to a non-empty string.

With this pull request, I now need to set only two instead of three settings (I no longer need to set the `bazel.lsp.args`). If we also remove `bazel.lsp.enabled`, we could cut down the setup effort to a single setting.

README.md
```diff
@@ -50,12 +50,14 @@ Clicking the "Stop" button in the debugger will kill the Bazel process being
 debugger, allowing you to halt the current build. The Bazel server, however,
 will continue running.
 
-## Using the LSP (Experimental)
+## Using a language server (Experimental)
 
-This extension can use [Facebook's starlark project](https://github.com/facebookexperimental/starlark-rust) as a language server.
+Some of the functionality of this extension, such as go to definition and completions, can be replaced with a language server. There are currently two in-progress implementations of this:
```
vogelsgesang, line 55: I wonder, if in the longterm we should aim for also replacing buildifier-based functionality (formatting, linting) with the language servers. As a part-time vim user, I personally would like that, because I can easily use a language server from vim, but I can't directly use buildifier. Also, it would probably simplify this extension a bit, if we have a single dependency (the language server) instead of two dependencies (the language server and buildifier).

(not really something to change in this pull request. Just a general idea)

withered-magic: rust-analyzer does something similar where `cargo check` (or `clippy`? forgot which one) is actually run by the language server itself and not the extension iirc

vogelsgesang: yes, `clangd` (for C++) also does something similar: They also embed the `clang-tidy` checks into `clangd` such that they can be displayed directly in the IDE. clangd does not actually call clang-tidy as a separate binary, though, but instead directly links against the clang-tidy library. Not sure how buildifier is built and if it could also be consumed as a library.

cameron-martin: rust-analyzer does forward the output of `cargo check`, but there is some doubt on whether this is the correct thing to. See https://github.com/rust-lang/rust-analyzer/issues/13437#issuecomment-1381758374.

package.json
cameron-martin: Yeah I think this makes sense.

README.md
withered-magic: But yeah I would probably advocate for some sort of middle ground where the language servers are the ones responsible for installing `buildifier` and running it on changes, instead of implementing that functionality themselves. It'd guarantee that we won't have any conflicts with any existing workflows that use `buildifier`, e.g. people running it in CI. (Otherwise if the language servers implemented `buildifier`'s functionality themselves, then to avoid said conflicts we'd have to make sure their behaviors match exactly which would be difficult to do).

withered-magic: Ooo interesting, I didn't see that thread!

cameron-martin: I can't comment on the security side, but I like the idea of this extension downloading tools. rust-analyzer used to do that before it came bundled with rust and the user experience was great.

vogelsgesang (approved):

README.md
vogelsgesang: I created #371 to track the idea of automatically bootstrapping the tools

withered-magic:
A bit off topic, but should we maybe add a disclaimer about certain functionalities not working correctly for either language server unless you're on a certain Bazel version? The main scenario I had in mind is for workspaces using `bzlmod`, since things like goto def won't work at all without the `bazel mod dump_repo_mappings` command, which wasn't added till `7.1.0` (pretty recently!).

jfirebaugh (approved):

cameron-martin:
@withered-magic sounds like a good idea, feel free to open a PR.

Issue #353, opened 2024-03-01
Feature request: shortcut to jump to BUILD or BUILD.bazel file


jiawen:
I primarily work on a mixed C++ and Python monorepo. I frequently find myself editing say a `.cc` file and wanting to jump to its corresponding `BUILD` file (in the same package). Is there a command to which I can assign a shortcut?

* Bonus points if we can jump to the "closest" target that has the current file in `srcs` or `hdrs`.
* Extra bonus points if there is a command to "jump to next target that has this file (the last focused source file) as a source attribute".

catskul:
I found this because I was just going to open the same feature request!

cameron-martin:
Looks like a duplicate of https://github.com/bazelbuild/vscode-bazel/issues/178

jiawen:
Oops you're right!

catskul:
Perhaps we should leave them both open until we get an official response from the core developers? They're slightly different requests. Not sure if that matters.

kchodorow:
I'd be interested in implementing both of these, if that's okay with the maintainers.

vogelsgesang:
Thanks for that offer and for your first Pull Request, @kchodorow 🚀 

I already replied in the pull request regarding the "go from a BUILD file to the referenced source file" functionality.
Regarding the "go from a source code file to corresponding BUILD target", a pull request would also be very welcome 🙂 

On a related note: I think having a "keep focus in sync with active editor" option for the Bazel Build Targets tree view would be pretty neat. I.e., whenever the currently active file in the editor is changed, the tree view is focused on the package of the currently edited file. Not sure if this is even possible with the APIs exposed by Visual Studio Code, though...

kchodorow:
Thanks! Source code -> build target is in progress (although probably dependent on the same StarLark/bazel query factors as the first PR).

> On a related note: I think having a "keep focus in sync with active editor" option for the Bazel Build Targets tree view would be pretty neat. I.e., whenever the currently active file in the editor is changed, the tree view is focused on the package of the currently edited file. Not sure if this is even possible with the APIs exposed by Visual Studio Code, though...

Interesting, that sounds useful! I'll take a look at what's available and let you know.

cbandera:
Hi @kchodorow are you still looking into this issue? Do you need help with bringing the implementation into a final PR?
I'd live to see this functionality in the extension.

cbandera:
I will try to look at it, but I'll probably need some time to get into the codebase and Typescript. 
[...]

PR #380, opened 2024-04-15
feat: Cmd-click to jump to a source file from a build rule

kchodorow:
Part of issue #353.

### DIFF
--- a/src/definition/bazel_goto_definition_provider.ts
+++ b/src/definition/bazel_goto_definition_provider.ts
@@ -24,6 +24,7 @@ import {
 import { Utils } from "vscode-uri";
 import { BazelQuery, BazelWorkspaceInfo, QueryLocation } from "../bazel";
 import { getDefaultBazelExecutablePath } from "../extension/configuration";
+import { blaze_query } from "../protos";
 
 // LABEL_REGEX matches label strings, e.g. @r//x/y/z:abc
 const LABEL_REGEX = /"((?:@\w+)?\/\/|(?:.+\/)?[^:]*(?::[^:]+)?)"/;
@@ -52,12 +53,18 @@ export class BazelGotoDefinitionProvider implements DefinitionProvider {
     const queryResult = await new BazelQuery(
       getDefaultBazelExecutablePath(),
       Utils.dirname(document.uri).fsPath,
-    ).queryTargets(`kind(rule, "${targetName}")`);
+    ).queryTargets(`kind(rule, "${targetName}") + kind(file, "${targetName}")`);
 
     if (!queryResult.target.length) {
       return null;
     }
-    const location = new QueryLocation(queryResult.target[0].rule.location);
-    return new Location(Uri.file(location.path), location.range);
+    const result = queryResult.target[0];
+    if (result.type == blaze_query.Target.Discriminator.RULE) {
+      const location = new QueryLocation(result.rule.location);
+      return new Location(Uri.file(location.path), location.range);
+    } else {
+      const location = new QueryLocation(result.sourceFile.location);
+      return new Location(Uri.file(location.path), new Position(0, 0));
+    }
   }
 }
--- a/src/definition/bazel_goto_definition_provider.ts
+++ b/src/definition/bazel_goto_definition_provider.ts
@@ -59,7 +59,7 @@ export class BazelGotoDefinitionProvider implements DefinitionProvider {
       return null;
     }
     const result = queryResult.target[0];
-    if (result.type == blaze_query.Target.Discriminator.RULE) {
+    if (result.type === blaze_query.Target.Discriminator.RULE) {
       const location = new QueryLocation(result.rule.location);
       return new Location(Uri.file(location.path), location.range);
     } else {


vogelsgesang:
Thanks for your contribution, @kchodorow! Much appreciated 🙂 

I am admittedly not sure which future direction we want to take regarding "Starlark / Bazel language features"... Should those be provided by the language server (which has a deeper logical understanding of the StarLark code) or by the TypeScript part of this extension (as proposed in this PR)?

On a related note, I am wondering if / how we would want to support `glob` expressions. I guess a hover information "glob expression matched 16 files" and a list of all the matched files would be useful? Not sure...

@cameron-martin @withered-magic What is your opinion on the long-term direction in which we should go here? Should this functionality be covered by the language servers or not?

Does one of your language servers already provide comparable functionality? If none of the language servers currently provides this functionality already, I would say we should move ahead with the PR as currently proposed (to unlock immediate user benefit) and then migrate this functionality into the language servers later on...

withered-magic:
There was actually a similar request for starpls here, but going in the opposite direction (from source file to corresponding `BUILD.bazel` files): https://github.com/withered-magic/starpls/issues/100. Haven't gotten a chance to look into too much but it actually should be pretty straightforward for to implement.

To me it seems reasonable for these sorts of functionalities to be covered by the language server and to have the language server be the only handler for fielding goto-def/find reference requests, which this feature sounds like

cameron-martin:
[bazel-lsp](https://github.com/cameron-martin/bazel-lsp) does have go-to-definition from BUILD files to source files. I'm generally for trying to push as much functionality into the language servers, but I'm also not against giving a several-line fix to functionality in this extension, particularly considering no language server is enabled by default.

vogelsgesang:
Works smoothly 🙂 

https://github.com/bazelbuild/vscode-bazel/assets/6820896/9ba609fb-0bb4-4252-8bae-7abd32b55700

One small issue which I noticed: The individual path components are linked individually (i.e. the `py` extension is linked, and not the complete file name). I think we can fix this by using a  [LocationLink](https://microsoft.github.io/monaco-editor/typedoc/interfaces/languages.LocationLink.html) instead of a [Location](https://microsoft.github.io/monaco-editor/typedoc/interfaces/languages.Location.html) and by setting the `originSelectionRange`. I think the necessary range should already be available from the variable `range` defined in line 43

Not introduced by you, but a patch to fix this would still be welcome :)

vogelsgesang:
> There was actually a similar request for starpls here, but going in the opposite direction (from source file to corresponding BUILD.bazel files): https://github.com/withered-magic/starpls/issues/100. Haven't gotten a chance to look into too much but it actually should be pretty straightforward to implement!
>
> To me it seems reasonable for these sorts of functionalities to be covered by the language server and to have the language server be the only handler for fielding goto-def/find reference requests, which this feature sounds like

Interesting idea... I didn't realize that we could also implement "go from source file to BUILD file" in the language server. So far, I was only considering the "go from BUILD file to source file" (i.e. this pull request here) in mind for the language server. I like the idea to implement both directions in the LS.

> [bazel-lsp](https://github.com/cameron-martin/bazel-lsp) does have go-to-definition from BUILD files to source files.

I just tried this out. Worked without any issues for me.

> I'm generally for trying to push as much functionality into the language servers, but I'm also not against giving a several-line fix to overlapping functionality in this extension, particularly considering no language server is enabled by default.

Then let's move ahead with this PR. I am still waiting if we can fix the `originSelectionRange`, and will then merge this PR.

withered-magic:
Just landed this on main for starpls! Will include it in the next release

kchodorow:
I was thinking it would just be a command palette option, since as you say there's no particular code we're "finding references" from.

@cameron-martin had the neat idea of having the Bazel Build Targets pane in the sidebar keep pace with whatever you currently are editing. I'm going to take a look at the API and see if that's feasible.

vogelsgesang:
> I was thinking it would just be a command palette option, since as you say there's no particular code we're "finding references" from.

Yes, that would also be my user interface of choice... Afaik, language servers can also contribute commands, but I never tried this before.

@kchodorow regarding this pull request at hand here: Are you still planning to improve `originSelectionRange` (see previous messages) or do you want me to merge this PR as is?

withered-magic:
Ah ok, I think my only reservation with making It a command palette option would be that other editors besides VSCode that are also using the LSP would also have to implement some custom way to use this functionality. Language servers can definitely support custom commands though! There's a [couple custom ones](https://github.com/withered-magic/starpls/blob/main/crates/starpls/src/extensions.rs) defined in starpls for debugging purposes.

kchodorow:
Oh, sorry, I missed the `originSelectionRange` comment! I'm a little confused about how to do what you're asking for... it doesn't seem like mouseover highlight is handled by this code. `provideDefinition` is triggered after the cmd/ctrl key is hit, AFAICT.

vogelsgesang:
ok, I will merge this PR and the see if can link the correct range in a follow-up commit

vogelsgesang (approved):

vogelsgesang:
see #382 for what I had in mind regarding `originSelectionRange`

PR #387, opened 2024-05-01
fix: Handle relative buildifier path without warning

ryanwilsonperkin:
Fixes https://github.com/bazelbuild/vscode-bazel/issues/386

Relative paths for buildifier are now supported as of v0.10.0, the warning should not be shown when the buildifier executable exists at the relative path specified.


ryanwilsonperkin:
(Haven't contributed previously or written VSCode extensions, would appreciate any feedback from the maintainers on how best to setup for testing this)

src/buildifier/buildifier_availability.ts
```diff
@@ -33,9 +35,15 @@ const BUILDTOOLS_RELEASES_URL =
...
 export async function checkBuildifierIsAvailable() {
   const buildifierExecutable = getDefaultBuildifierExecutablePath();
+
   // Check if the program exists (in case it's an actual executable and not
   // an target name starting with `@`).
-  if (!buildifierExecutable.startsWith("@")) {
+  const isTarget = buildifierExecutable.startsWith("@")
+
+  // Check if the program exists as a relative path of the workspace
+  const isRelative = fs.existsSync(path.join(vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath, buildifierExecutable));
```
vogelsgesang, line 44: The `existsSync` / `exists` APIs are deprecated in node. Can we switch to `fs.stat` instead?

vogelsgesang, line 44: Please use async APIs instead of the sync API. You should be able to use `util.promisify` to get a promise-based variant of `fs.exists` which you can then `await` on. See https://github.com/bazelbuild/vscode-bazel/blob/8972b0eb8270efe471245f5846efb335fd760fda/src/bazel/bazel_info.ts#L20 for an example

vogelsgesang, line 44: What happens for `@` paths here? `@` is not a valid character in file system paths. Does `fs.exists` gracefully return `false` for such file paths or are we getting some kind of exception?

(Maybe just move the complete `fs.existsSync` check into the `if(!isTarget)`, just to be on the safe side)

vogelsgesang, line 44: Afaict, `isRelative` will also be true for absolute paths, if that path is found? Maybe rename this to `foundWithoutPathEnv` or something similar?

vogelsgesang (commented):
Thanks for your contribution and for the bug report!

A couple of stylistic issues, but on a high-level you found the right place to fix this 🙂 

Regarding testing: We currently don't have a good testing strategy, yet. In #366, I am about to introduce the first unit tests, but we are still far away from a satisfactory testing solution, and I don't want to put it on you to build that infrastructure for the overall project. Hence, I would be fine merging this without test cases

vogelsgesang:
I just saw that our CI doesn't like the code formatting...
You can run `npm run format-fix` to fix the formatting

src/buildifier/buildifier_availability.ts
ryanwilsonperkin: `exists` is deprecated but `existsSync` isn't. I'm happy to change it to the async version though per your other comment 👍🏻

ryanwilsonperkin: Since we'll do a check if the file exists and it isn't a valid path name it'll just set the `isRelative` (now `pathExists`) to false

ryanwilsonperkin: Good call, renamed

ryanwilsonperkin:
Thanks for the feedback @vogelsgesang ! Pushed some fixes

vogelsgesang (approved):

vogelsgesang:
Looks good to me, thanks again!

vogelsgesang:
The next release to the Marketplace will probably still be a while. In the meantime, you can get a prebuilt pre-release version from the [GitHub Action artifacts](https://github.com/bazelbuild/vscode-bazel/actions/runs/8931739553)

PR #388, opened 2024-05-02
feat(coverage): Demangling of Java function names in coverage results

vogelsgesang:
This commit adds demangling support for Java. I couldn't find any good
third-party library for this, so I quickly implemented my own, based on
https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2

It might still be have a couple of bugs, but at least for my simple
smoke tests, demangling seems to work just fine.

Works towards #362


vogelsgesang:
Gentle ping 🙂

test/lcov_parser.test.ts
```diff
@@ -96,13 +96,12 @@ describe("The lcov parser", () => {
...
       assert(initFunc !== undefined);
-      assert.equal(initFunc.name, "com/example/myproject/Greeter::<init> ()V");
+      assert.equal(initFunc.name, "void Greeter::<init>()");
       assert.equal(initFunc.executed, 1);
       const convertFunc = getFunctionByLine(fileCov, 28);
       assert.equal(
         convertFunc.name,
-        "com/example/myproject/Greeter::convertStreamToString " +
-          "(Ljava/io/InputStream;)Ljava/lang/String;",
+        "String Greeter::convertStreamToString(InputStream)",
```
cameron-martin, line 104: Do we definitely want to drop the namespace here, and only keep the type name?

vogelsgesang: I think it makes sense to drop the prefix, yes.

In the VSCode UI, the functions are grouped by file paths. See https://code.visualstudio.com/updates/v1_88#_test-coverage for a screen shot. Note how in the lower left part, you have a file system tree. If you expand one of those file names, you can see all functions inside that file.

As such, function names only need to be unique in the context of a single file. Repeating the package name in each and every function name would not add any additional information, but would clutter the UI

cameron-martin (approved):

test/lcov_parser.test.ts
cameron-martin: Makes sense, thanks!

PR #393, opened 2024-05-06
chore: Remove unmaintained `CONTRIBUTORS` and `AUTHORS` files

vogelsgesang:
Those two files were not maintained for a long time, and I don't know
what value they are providing. Hence, I am deleting them


vogelsgesang:
@meteorcloudy do you have any background on why those files exist? Do we maybe need to keep them due to some Google-internal compliance reasons?

meteorcloudy (approved):

meteorcloudy:
I don't think there is any internal policy for this. Let's update to reflect the reality.

cameron-martin (approved):

PR #398, opened 2024-06-07
feat(coverage): Demangling C++ & Rust function names in coverage results

vogelsgesang:
Demangle C++ and Rust symbols using `c++filt` and `rustfilt`. We rely on those tools to be installed on the `$PATH`.

I was also considering whether I should compile `c++filt` and / or `rustfilt` to WebAssembly, but decided against it, given that this would make the build much more complicated and it is currently unclear to me whether we will stick with the "traditional" tool chains (npm, webpack, cargo, ...) or if we will pivot to using Bazel.

### DIFF
--- a/src/test-explorer/lcov_parser.ts
+++ b/src/test-explorer/lcov_parser.ts
@@ -1,6 +1,11 @@
 import * as vscode from "vscode";
 import * as path from "path";
+import * as child_process from "child_process";
+import * as which from "which";
+import * as util from "util";
 import { assert } from "../assert";
+
+const execFile = util.promisify(child_process.execFile);
 
 /**
  * Demangle JVM method names.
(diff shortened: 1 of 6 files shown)


src/test-explorer/lcov_parser.ts
```diff
@@ -218,17 +239,19 @@ export function parseLcov(
...
-            const demangled = demangleJVMMethodName(funcName) ?? funcName;
+            // Demangle the name.
+            // We must first try rustfilt before trying c++filt.
+            // The Rust name mangling scheme is intentionally compatible with
+            // C++ mangling. Hence, c++filt will be succesful on Rust's mangled
+            // names. But rustfilt provides more readable demanglings, and hence
+            // we prefer rustfilt over c++filt. For C++ mangled names, rustfilt
+            // will fail and we will fallback to c++filt.
+            // See https://internals.rust-lang.org/t/symbol-mangling-of-rust-vs-c/7222
+            const demangled =
```
cameron-martin, line 250: It's a shame that we don't know what language the lcov was generated from here. Is there no reasonable way of getting this info out of Bazel?

vogelsgesang: The LCOV file from Bazel might even be merged across multiple different languages. E.g., if I run `./bazel coverage //...`, Bazel will merge coverage between Java, C++ and Go test cases.

We could use some heuristic by looking at the file extension, but then again: What would we gain by doing so?

What is your underlying concern with trying multiple decoders and using the first one that can interpret the name?

cameron-martin: I was just concerned that a decoder could incorrectly demangle a symbol from another language. However, I'm not sure how likely this is and if this isn't possible to easily determine the language then I'm happy with this approach.

cameron-martin (approved):

src/test-explorer/lcov_parser.ts
vogelsgesang: I think we should be good here in almost all cases. As soon as we run into it, we will find a better solution :slightly_smiling_face:

PR #404, opened 2024-06-29
feat: Extend CodeLens to build and run more targets

cramertj:
Previously, `_binary` targets could only be run, `_test` taregets could only be tested, and all other targets could only be built. This change allows all targets to be built and all non-`_library` targets to be run.

Additionally, this change shortens the names of the targets in codelens, as otherwise the additional commands caused a lot of clutter: `Run //foo/bar/baz:flubber  Build //foo/bar/baz:flubber` etc.

Example after change:
bazel-vscode-screenshot

### DIFF
--- a/src/codelens/bazel_build_code_lens_provider.ts
+++ b/src/codelens/bazel_build_code_lens_provider.ts
@@ -20,6 +20,25 @@ import { getDefaultBazelExecutablePath } from "../extension/configuration";
 import { blaze_query } from "../protos";
 import { CodeLensCommandAdapter } from "./code_lens_command_adapter";
 
+/** Computes the shortened name of a Bazel target.
+ *
+ * For example, if the target name starts with `//foo/bar/baz:fizbuzz`,
+ * the target's short name will be `fizzbuzz`.
+ *
+ * This allows our code lens suggestions to avoid filling users' screen with
+ * redundant path information.
+ *
+ * @param targetName The unshortened name of the target.
+ * @returns The shortened name of the target.
+ */
+function getTargetShortName(targetName: string): string {
+  const colonFragments = targetName.split(":");
+  if (colonFragments.length !== 2) {
+    return targetName;
+  }
+  return colonFragments[1];
+}
+
 /** Provids CodeLenses for targets in Bazel BUILD files. */
 export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
   public onDidChangeCodeLenses: vscode.Event<void>;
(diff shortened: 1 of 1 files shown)


src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -20,6 +21,48 @@ import { getDefaultBazelExecutablePath } from "../extension/configuration";
...
+ * This allows our code lens suggestions to avoid filling users' screen with
+ * redundant path information.
+ *
+ * @param targetName The unshortened name of the target.
+ * @param bazelWorkspaceInfo The currrent Bazel workspace.
+ * @param location The location of the target's BUILD.bazel file.
+ *
+ * @returns The shortened name of the target.
+ */
+function getTargetShortName(
```
cameron-martin, line 53: Can't we always strip the package here, since targets we're considering will always be in the current package?

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -107,40 +150,62 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
+        commands.push({
+          commandString: "bazel.testTarget",
+          name: "Test",
+        });
+      }
+
+      // Targets which are not libraries may support running.
+      //
+      // Without checking the Bazel rule's `executable` attribute we can't know
+      // for sure which targets can be run.
```
cameron-martin, line 181: This can be determined using a `cquery`, but this is quite a large change. Considering we used a heuristic before, I think this is fine.

```
bazel cquery '//:all' --output=starlark --starlark:expr "json.encode({'target': target.label.name, 'executable': providers(target)['FilesToRunProvider'].executable != None})"
```

cramertj: Yeah, I wasn't sure if this was worth the cost of the added query + latency. I'm happy to leave it as-is or make this change-- let me know which you'd prefer.

src/codelens/bazel_build_code_lens_provider.ts
cramertj: I wasn't sure if this was true (whether target rule names are always within the current package, or whether it's possible to declare something outside of the current package). I can't think of a way you'd break this, though. Changed.

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -20,6 +21,40 @@ import { getDefaultBazelExecutablePath } from "../extension/configuration";
...
 
+/** Returns a target string without some prefix.
+ *
+ * If the target string does not have the prefix, returns the string unchanged.
+ *
+ * @param target The string whose prefix to remove.
+ * @param prefix The prefix to remove, if present.
+ * @returns The target string without the provided prefix.
+ */
+function withoutPrefix(target: string, prefix: string): string {
```
cameron-martin, line 32: Isn't this unused now?

src/codelens/bazel_build_code_lens_provider.ts
cameron-martin: I'm happy to leave it as-is, particularly because `cquery` has extra cost due to running the analysis phase.

cameron-martin:
Can you fix the lint errors too?

cameron-martin (approved):

romannikov:
1. Could you please hide repeated target names?
2. Could you please add a command to copy a target to a clipboard? command is already defined. It'll look like as a first command in the commands array:
```
commands.push({
  commandString: "bazel.copyTargetToClipboard",
  name: "Copy",
});
```
Many thanks!

cameron-martin:
@romannikov These sound like good changes, but also quite separate from this change. I'd be happy to accept these changes in separate PRs though.

romannikov:
> @romannikov These sound like good changes, but also quite separate from this change. I'd be happy to accept these changes in separate PRs though.

I'll happy to send these changes in a follow up pull request, since this one will be committed.

vogelsgesang (approved):
Thanks for the pull request! Nice improvement 🙂

PR #406, opened 2024-07-03
refactor: Move Bazel wrapper commands into separate file

vogelsgesang:
This in preparation of adding a new "Measure Coverage" command, but I don't want to make the main `extension.ts` file even longer. Hence, I first move all the commands which simply wrap underlying Bazel commands into a separate file.


src/extension/extension.ts
```diff
@@ -99,24 +92,7 @@ export async function activate(context: vscode.ExtensionContext) {
...
-    vscode.commands.registerCommand("bazel.runTarget", bazelRunTarget),
-    vscode.commands.registerCommand("bazel.testTarget", bazelTestTarget),
-    vscode.commands.registerCommand("bazel.testAll", bazelTestAll),
-    vscode.commands.registerCommand(
-      "bazel.testAllRecursive",
-      bazelTestAllRecursive,
-    ),
-    vscode.commands.registerCommand("bazel.clean", bazelClean),
+    ...activateCommands(),
     vscode.commands.registerCommand("bazel.refreshBazelBuildTargets", () => {
```
cameron-martin, line 96: How come some of the commands are still in extension.ts?

vogelsgesang: My underlying thinking was:

The commands which I moved out into an own file are direct Bazel wrapper commands.
The `refreshBazelBuildTargets` and `copyTargetToClipboard` are not wrappers around Bazel commands, but are UI-specific commands.

I obviously failed to communicate this, both in the commit message, the file name and the function name.

I see two options:

**Option 1**:
* rename the file `commands.ts` to `bazel_wrapper_commands.ts`
* rename the file `activateCommands` -> `activateBazelWrapperCommands`
* update the commit message accordingly

**Option 2**
* also move `copyTargetToClipboard` and `refreshBazelBuildTargets` over

I slightly tend to option 1, because I think we should structure the code based on "features" (wrapping the Bazel commands; providing a tree view; providing code lenses; ...) and not "implementation vehicle" (commands, treeviews, ...). Do you agree with that?

cameron-martin: Makes sense - I agree that organising the code around features is better, so I'm happy with option 1.

vogelsgesang: I finally updated this Pull Request 🎉 (took me longer than I hoped for. Too many other things going on...)

cameron-martin (approved):

PR #417, opened 2024-10-15
fix(coverage): Do not emit error on BRDA:0 entries

cwahbong:
Coverage report generators in Coverage.py would generate BRDA:0 for exit branches.

Ref: https://github.com/nedbat/coveragepy/blob/5467e1f/coverage/lcovreport.py#L108-L112

Also small string fixes for error messages in BRDA.

Improves #362.


src/test-explorer/lcov_parser.ts
vogelsgesang, line 1: could you add a test case for this?
(Maybe update one of the existing test cases accordingly instead of adding a new one from scratch)

cwahbong: Let me add several manual crafted invalid entries instead of generating one from source.

cwahbong:
I just saw other discussions about the misuse of the line number fields. Is it making sense for being tolerant the negative numbers or simply drop the invalid entries and maybe put a warning message somewhere, either pop up or on output window (I did not find the output window like in other plugin for this one) in case other coverage tools also misuse the line number field?

* https://github.com/nedbat/coveragepy/issues/1846
* https://github.com/linux-test-project/lcov/blob/master/man/geninfo.1#L1438

src/test-explorer/lcov_parser.ts
cwahbong: Done, PTAL

vogelsgesang (approved):

vogelsgesang:
> Is it making sense for being tolerant the negative numbers or simply drop the invalid entries and maybe put a warning message somewhere

Maybe. Let's keep it as is, and then potentially ignore also negative numbers in case we actually get reports of this being a real-world issue. (I prefer "fail early, fail hard" since this makes it much simpler to find potential issues/unexpected behavior).

Issue #416, opened 2024-10-14
Handle external dependencies


cwahbong:
Seems the current plugin implementation does not handle the external dependencies well: when specify "Query Expression" in configuration like: `@repo//...`, the tree explorer, coverage, and the quick pick is not showing the dependencies from external dependencies. This makes some workspaces that's assembling local repository rules not working.

I already have several local workaround fixes for this, but it would be better to also append the test cases for better coverage and prevent regressions for existing users..

- [ ] Coverage: handle the generated report path's `SF:external/<repo>/....`  entries.
- [ ] Ｑuick picks: replace hard coded queries to query expression.
- [ ] Tree explorer:
  - [ ] Replace hard coded queries to query expression.
  - [ ] Make the grouping handles the package name starting with `@<repo>//`.

cbandera:
## Scoping this issue: two independent problem axes, and what "external" means today

While fixing #696 (relative-label resolution for the main repo) we ran into the same underlying gap this issue describes, but found the problem is broader than the checklist above covers. Posting this as a consolidated problem-space writeup so whoever picks this up next doesn't have to re-derive it.

### What counts as a "local directory considered external"

Two distinct Bazel mechanisms produce the same situation — a directory physically nested inside (or beside) the main workspace's directory tree, editable like any other package, but which Bazel treats as a **separate repository** with its own label namespace:

- **Legacy WORKSPACE**: `local_repository(name = "repo", path = "...")` — this issue's original report.
- **Bzlmod (MODULE.bazel)**: `local_path_override(module_name = "mod", path = "...")` — not mentioned above, presumably because this issue predates this repo's bzlmod support (#684/#687). Same shape, same bug.

In both cases, any label inside that directory must be `@repo//...` (or the canonical `@@repo+//...` form) — never bare `//...` — even though the file sits right next to ordinary in-repo packages on disk.

**This is the same physical layout as #684's scenario S5/S4** (nested `MODULE.bazel` confusing root auto-detection), but a **different bug**. #687's `bazel.workspacePath` fixes *which directory Bazel is rooted at* for queries; it does **not** fix *which label namespace a file's labels belong to*. A file inside a `local_path_override`'d module can have a perfectly-resolved, pinned workspace root and still get the wrong label. Worth stating explicitly since it's easy to assume #687 already covered this (#684 itself flags this as scenario "S6" and explicitly defers it here, but this issue never linked back).

### Axis A — Query construction (what this issue's checklist already tracks)

[...]

PR #418, opened 2024-10-15
fix(coverage): Expand external local repo in coverage report

cwahbong:
Rules under repo local_repository() are external repo but symlinked to a local path, usually in the same workspace.

No matter if the local path is in the same workspace, the users should usually want to check the local path specified in local_repository rules instead of the path in bazel output
(`<bazel workspace>/bazel-<repo>/external/...`) in
file explorer.

Works towards #416. Improves #362.

### DIFF
--- a/src/bazel/tasks.ts
+++ b/src/bazel/tasks.ts
@@ -156,10 +156,12 @@ async function onTaskProcessEnd(event: vscode.TaskProcessEndEvent) {
   if (taskDefinition.command === "coverage" && rawExitCode === 0) {
     // Find the coverage file and load it.
     const workspaceInfo = await getWorkspaceInfoFromTask(task.scope);
-    const outputPath = await new BazelInfo(
+    const bazelInfo = new BazelInfo(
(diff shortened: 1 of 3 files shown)


src/bazel/tasks.ts
```diff
@@ -183,12 +185,10 @@ async function onTaskProcessEnd(event: vscode.TaskProcessEndEvent) {
...
         );
       } else {
-        // Show the coverage date
-        await showLcovCoverage(
-          description,
-          workspaceInfo.bazelWorkspacePath,
-          covFileStr,
-        );
+        // The `bazel coverage` runs the build/test/coverage in sandboxes with
+        // the similar/same layout of execution root, thus making it as the base
```
vogelsgesang, line 189:
```suggestion
        // the similar/same layout of execution root, thus using it as the base
```

vogelsgesang (approved):
LGTM. Thanks for the contribution!

Waiting before merging this, to also give others a chance to review

cameron-martin (approved):

PR #420, opened 2024-10-20
refactor(workspace-tree): Refine event and dispose handling

cwahbong:
Less dependency for BazelWorkspaceTreeProvider, making the future testing changes simpler.

Also small clean ups for this capturing and making the styles (naming and initializations) like
https://github.com/microsoft/vscode-extension-samples/blob/2f83557/tree-view-sample/src/nodeDependencies.ts

### DIFF
--- a/src/workspace-tree/bazel_workspace_tree_provider.ts
+++ b/src/workspace-tree/bazel_workspace_tree_provider.ts
@@ -25,9 +25,9 @@ export class BazelWorkspaceTreeProvider
   implements vscode.TreeDataProvider<IBazelTreeItem>, vscode.Disposable
 {
   /** Fired when BUILD files change in the workspace. */
-  private readonly _onDidChangeTreeData =
+  private readonly onDidChangeTreeDataEmitter =
     new vscode.EventEmitter<IBazelTreeItem | void>();
-  public readonly onDidChangeTreeData = this._onDidChangeTreeData.event;
+  public readonly onDidChangeTreeData = this.onDidChangeTreeDataEmitter.event;
 
   /** The cached toplevel items. */
   private workspaceFolderTreeItems: BazelWorkspaceFolderTreeItem[] | undefined;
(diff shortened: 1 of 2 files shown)


src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -22,49 +22,38 @@ import { BazelWorkspaceFolderTreeItem } from "./bazel_workspace_folder_tree_item
...
  */
 export class BazelWorkspaceTreeProvider
-  implements vscode.TreeDataProvider<IBazelTreeItem>
+  implements vscode.TreeDataProvider<IBazelTreeItem>, vscode.Disposable
 {
-  public onDidChangeTreeData: vscode.Event<IBazelTreeItem | void>;
-
   /** Fired when BUILD files change in the workspace. */
-  private onDidChangeTreeDataEmitter =
+  private readonly _onDidChangeTreeData =
```
vogelsgesang, line 28: Distinguishing between the emitter and the even only by the `_` is a bit implicit. Let's rather make this more explicit in the variable name

```suggestion
  private readonly onDidChangeTreeDataEmitter =
```

vogelsgesang (approved):

src/workspace-tree/bazel_workspace_tree_provider.ts
cwahbong: Edited all occurrences in this file.

cwahbong:
The rest of work for [workspace handling external dependency](https://github.com/cwahbong/vscode-bazel/tree/workspace-tree) is still without test cases (it-works-on-my-machines), thus uploading only a change without much behavior change first.

vogelsgesang (approved):

vogelsgesang:
> The rest of work for [workspace handling external dependency](https://github.com/cwahbong/vscode-bazel/tree/workspace-tree) is still without test cases (it-works-on-my-machines), thus uploading only a change without much behavior change first.

Thanks for the context! Sounds good. Even independent of any future changes, this is definitely an improvement. Great work, keep it going 🙂 

By the way: You can reach me and the other maintainers of this extension on https://slack.bazel.build/, in case you want to discuss architecture / code design / ...

PR #421, opened 2024-10-20
Bundle the extension using esbuild

cwahbong:
This should not affect the existing build and test flow, but adding an additional step right before building vsix package.

The esbuild.js is copied from VSCode extension generator template: github.com/microsoft/vscode-generator-code/tree/4812b7e/generators/app/templates/ext-command-ts/vscode-esbuild

This reduces the package size from ~5MB -> ~250KB.


.vscodeignore
```diff
@@ -1,7 +1,9 @@
+.github/**
```
vogelsgesang, line 1: Potential improvement:

In other projects, I usually [copy over](https://github.com/salesforce-misc/bazelrc-lsp/blob/aaf7ed6662c8d426449a92687a40c24737fc5181/vscode-extension/build.js#L22) the relevant files into the `dist` folder and then [call vsce on that subfolder](https://github.com/salesforce-misc/bazelrc-lsp/blob/aaf7ed6662c8d426449a92687a40c24737fc5181/vscode-extension/package.json#L46)

Thereby, I have to opt-in each included file, instead of having to opt-out via the `.vscodeignore` file, thereby avoiding unintended bloat in the packaged extension

Let me know what you think :)

package.json
```diff
@@ -467,7 +467,7 @@
         "test": "./scripts/test.sh",
         "package": "vsce package",
         "update-snapshot": "./scripts/test.sh -u",
-        "vscode:prepublish": "./scripts/build.sh",
+        "vscode:prepublish": "./scripts/build.sh && node esbuild.js",
```
vogelsgesang, line 470: shouldn't this use the `--production` flag? Does this minify the bundle even further?

```suggestion
        "vscode:prepublish": "./scripts/build.sh && node esbuild.js --production",
```

vogelsgesang, line 470: Maybe we want to merge `build.sh` and `esbuild.js` into a single `build.js`? That would potentially keep our buld more straightforward

package.json
vogelsgesang, line 1: > This should not affect the existing build and test flow, but adding an additional step right before building vsix package.

I would not be opposed to actually add esbuild also to our non-release build and test flow. In my experience, esbuild is fast enough.

WDYT, @cameron-martin?

vogelsgesang (commented):
Thanks for this PR! Properly bundling and minifying the extensions has been on my TODO list for a long time already.

I know that there were discussions on "should we build this extension via Bazel", but afaik they stalled. As such, I would be happy to move on with this PR. We can later still switch to Bazel, in case somebody is willing to put in the effort for this switch

package.json
cwahbong: I was a little bit worried about the debugging flow (like the breakpoint) would be affected if we add esbuilt. We'd better add them into debug (non-release?) build after we figure this out.

package.json
cwahbong: I just made a package.sh for doing the dependent tasks instead of doing it in prepublish. To me prepublish is some chore after build package and just befre the publishing (like login, checking credentials).

.vscodeignore
cwahbong: SG. I've also removed `.vscodeignore`.

scripts/package.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Bazel Authors. All rights reserved.
```
vogelsgesang, line 3:
```suggestion
# Copyright 2024 The Bazel Authors. All rights reserved.
```

vogelsgesang (approved):
Looks good to me. I am still waiting before merging this, to also give others a chance to review

package.json
vogelsgesang: I would hope that debugging should still work thanks to the source-maps (for debugging, we would call eslint without the `--production` flag, of course). I made good experiences with that when debugging webpages. However, I never tested if this actually works when debugging VS-Code extensions.

Long story short: I am optimistic that we should be able to use `esbuild.sh` also for the local debugging flow. But I am fine if we take the pessimistic approach, until it's actually proven that the optimistic approach works 🙂

cwahbong:
> Thanks for this PR! Properly bundling and minifying the extensions has been on my TODO list for a long time already.
> 
> I know that there were discussions on "should we build this extension via Bazel", but afaik they stalled. As such, I would be happy to move on with this PR. We can later still switch to Bazel, in case somebody is willing to put in the effort for this switch

I did a [quick try (still not working)](https://github.com/cwahbong/vscode-bazel/tree/bazelisk) for using Bazel build for building this extension. It is more complicated even for only making a `pbts` command works in the build system. At first I thought it was at most just a bunch of `genrule`s but it turns out it would be making `npm` repositories and I don't think I can finish them in a short time. (comparing to this pull request the first version is just done in under 10 minutes...)

package.json
cwahbong: I just made the corresponding changes in `launch.json` to support the debugging. I only validate that local breakpoint could be set in either "Launch Extension" and "Launch Extension Tests".

cwahbong:
I just made the corresponding changes in launch.json to support the debugging. I only validate that local breakpoint could be set in either "Launch Extension" and "Launch Extension Tests".

However I don't think I'm really understand how / why Debug Server works. It seems to be a debugging service for bazel build file instead of the plugin or typescript/javascript stuffs. I just made them into another binary, updating the referencedin `package.json` as well.

.vscode/launch.json
```diff
@@ -6,28 +6,29 @@
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
-            "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
-            "stopOnEntry": false,
```
cwahbong, line 10: stopOnEntry is reported invalid entry in my vscode, thus removing it.

.vscode/launch.json
```diff
@@ -6,28 +6,29 @@
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
-            "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
```
cwahbong, line 9: workspaceRoot is now [deprecated](https://code.visualstudio.com/docs/editor/variables-reference#_why-isnt-workspaceroot-documented)

.vscodeignore
```diff
@@ -0,0 +1,10 @@
+# Ignore all files by default.
```
cwahbong, line 1: I found some examples and practices using `.vscodeignore` for white listing: so that we can invoke the `vsce` without changing the working directory or copying the included resources, and the layout for the resources should be the same no matter we minified or not.

cwahbong:
Still looking into the "watch" flow. I also have no idea for how to test these flow automatically to prevent regression.
For now I assume all developers are daily using it and would be triggered once broken again.

package.json
```diff
@@ -483,6 +488,7 @@
         "eslint-plugin-jsdoc": "^48.2.2",
         "js-yaml": "^4.1.0",
         "mocha": "^10.4.0",
+        "npm-run-all": "^4.1.5",
         "prettier": "^3.2.5",
```
cwahbong, line 492: This is referenced from https://github.com/microsoft/vscode-generator-code/blob/4812b7e/generators/app/templates/ext-command-ts/vscode-esbuild/package.json, but it has 1700+ dependent packages...

adam-azarchs: That does seem like a lot of extra dependency weight (with the attendant impact on CI times and security alert noise) for something of marginal utility.  Developers need to take explicit action to reload the extension anyway.

scripts/build.sh
```diff
@@ -19,20 +19,8 @@ set -eu
...
-fi
+node esbuild-debug-adapter.js
+node esbuild.js
 
-# Convert yaml language definition to json form requred by vscode.
-js-yaml syntaxes/bazelrc.tmLanguage.yaml > syntaxes/bazelrc.tmLanguage.json
-
-# Compile the rest of the project.
+# Compile the rest of the project for development flow like testing.
 tsc "$@" -p ./
```
adam-azarchs, line 26: Why do you still need `tsc` here?  `esbuild` handles compilation.  Running `tsc` remains useful for testing but not so much for build.

.vscodeignore
```diff
@@ -1,18 +1,10 @@
...
-CODEOWNERS
-CONTRIBUTORS
-tsconfig.json
+!README.md
+!LICENSE
+!package.json
+!dist/**/*.js
+!icons/**/*.+(svg|license)
+!media/**/*
+!syntaxes/**/*.+(json|yaml|license)
```
adam-azarchs, line 10: yaml file shouldn't be included in the extension, only the form that's been compiled to json.
```suggestion
!syntaxes/**/*.+(json|license)
```
Do note that it was previously excluded.

PR #424, opened 2024-10-24
refactor(workspace-tree): Future proof icon path resolution

cwahbong:
Current resource path resolution is using the source file as the base directory, this would fail if we minify the package into single js.

Use extension path as the base directory is more reliable: it is also working for the current packaging, and we just have to make sure the resource layout of the minified package remains the same.

An enum is provided with a test looping through all value and checking the existence of the icon files in extension.

### DIFF
--- a/src/workspace-tree/bazel_package_tree_item.ts
+++ b/src/workspace-tree/bazel_package_tree_item.ts
@@ -78,8 +78,8 @@ export class BazelPackageTreeItem
     return this.packagePath.substring(this.parentPackagePath.length + 1);
   }
 
-  public getIcon(): vscode.ThemeIcon {
-    return vscode.ThemeIcon.Folder;
+  public getIconName(): undefined {
+    return undefined;
   }
 
   public getTooltip(): string {
--- a/src/workspace-tree/bazel_package_tree_item.ts
+++ b/src/workspace-tree/bazel_package_tree_item.ts
@@ -23,6 +23,7 @@ import { getDefaultBazelExecutablePath } from "../extension/configuration";
 import { blaze_query } from "../protos";
 import { BazelTargetTreeItem } from "./bazel_target_tree_item";
 import { IBazelTreeItem } from "./bazel_tree_item";
+import { Resources } from "../extension/resources";
 
 /** A tree item representing a build package. */
 export class BazelPackageTreeItem
@@ -44,6 +45,7 @@ export class BazelPackageTreeItem
    * {@code packagePath} should be stripped for the item's label.
    */
   constructor(
+    private readonly resources: Resources,
     private readonly workspaceInfo: BazelWorkspaceInfo,
     private readonly packagePath: string,
     private readonly parentPackagePath: string,
@@ -62,7 +64,11 @@ export class BazelPackageTreeItem
       sortByRuleName: true,
     });
     const targets = queryResult.target.map((target: blaze_query.ITarget) => {
-      return new BazelTargetTreeItem(this.workspaceInfo, target);
+      return new BazelTargetTreeItem(
+        this.resources,
+        this.workspaceInfo,
+        target,
+      );
     });
     return (this.directSubpackages as IBazelTreeItem[]).concat(targets);
   }
@@ -78,8 +84,8 @@ export class BazelPackageTreeItem
     return this.packagePath.substring(this.parentPackagePath.length + 1);
   }
 
-  public getIconName(): undefined {
-    return undefined;
+  public getIcon(): vscode.ThemeIcon {
+    return vscode.ThemeIcon.Folder;
   }
 
   public getTooltip(): string {
--- a/src/workspace-tree/bazel_workspace_tree_provider.ts
+++ b/src/workspace-tree/bazel_workspace_tree_provider.ts
@@ -96,18 +96,6 @@ export class BazelWorkspaceTreeProvider
     return Promise.resolve([]);
   }
 
-  private getIcon(element: IBazelTreeItem): string | vscode.ThemeIcon {
-    const iconName = element.getIconName();
-    if (iconName === undefined) {
-      if (element.mightHaveChildren()) {
-        return vscode.ThemeIcon.Folder;
-      } else {
-        return vscode.ThemeIcon.File;
-      }
-    }
-    return this.resources.getIconPath(iconName);
-  }
-
   public getTreeItem(element: IBazelTreeItem): vscode.TreeItem {
     const label = element.getLabel();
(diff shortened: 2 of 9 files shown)


cwahbong:
This fixes the icon issues happened in #421.

src/workspace-tree/bazel_package_tree_item.ts
```diff
@@ -78,8 +78,8 @@ export class BazelPackageTreeItem
     return this.packagePath.substring(this.parentPackagePath.length + 1);
   }
 
-  public getIcon(): vscode.ThemeIcon {
-    return vscode.ThemeIcon.Folder;
+  public getIconName(): undefined {
+    return undefined;
   }
```
vogelsgesang, line 83: why?

src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -34,12 +35,20 @@ export class BazelWorkspaceTreeProvider
 
   private disposables: vscode.Disposable[] = [];
 
+  public static fromExtensionContext(
```
vogelsgesang, line 38: afaict, we could just add a `context: vscode.ExtensionContext` parameter to the constructor? Not sure why we would need an additional factory function

src/workspace-tree/icons.ts
```diff
@@ -60,23 +59,19 @@ const SPECIFIC_RULE_CLASS_ICONS: Record<string, string> = {
...
 export function getBazelRuleIcon(
   target: blaze_query.ITarget,
-): string | vscode.ThemeIcon {
+): IconName | undefined {
   const ruleClass = target.rule.ruleClass;
   let iconName = SPECIFIC_RULE_CLASS_ICONS[ruleClass];
   if (!iconName) {
     if (ruleClass.endsWith("_binary")) {
-      iconName = "binary";
+      iconName = IconName.BINARY;
```
vogelsgesang, line 67: random idea: maybe we should make this lookup configurable? I assume many people have custom in-house Bazel rules, and still might want to have nice icons for those

src/workspace-tree/icons.ts
```diff
@@ -60,23 +59,19 @@ const SPECIFIC_RULE_CLASS_ICONS: Record<string, string> = {
  */
 export function getBazelRuleIcon(
   target: blaze_query.ITarget,
-): string | vscode.ThemeIcon {
+): IconName | undefined {
```
vogelsgesang, line 62: just a word of caution: `strictNullChecks` is currently disabled in our `tsconfig.json`. So the Typescript compiler doesn't actually check those type annotations closely.

This is still legacy, we didn't find time to clean up the code such that we could enable `strictNullChecks` , yet

(I am not asking you to implement this as part of this commit - just as food for thought 🙂)

test/resources.test.ts
```diff
@@ -0,0 +1,17 @@
+import assert = require("assert");
+import { IconName, Resources } from "../src/extension/resources";
+import * as fs from "fs/promises";
+import * as vscode from "vscode";
+
+describe("The resources", () => {
+  const resources = new Resources(
+    vscode.extensions.getExtension("BazelBuild.vscode-bazel").extensionPath,
+  );
+  it("provides valid icon paths", async () => {
```
vogelsgesang, line 10: for this test case to provide the intended test coverage, wouldn't we need to run the tests against the bundled code? Which would bring us back to https://github.com/bazel-contrib/vscode-bazel/pull/421#discussion_r1807583031

src/workspace-tree/icons.ts
cwahbong: SG, and I'm also thinking about if we should try adopt [vscode-icons](https://github.com/vscode-icons/vscode-icons/wiki/ListOfFiles) in the future, but using same set of the icons may also confuse the users between "files view" and "workspace rules view"...

src/workspace-tree/bazel_workspace_tree_provider.ts
cwahbong: You're right. The main intention is that in later change for making the `BazelWorkspaceTreeProvider` easier to be tested, I'm also pulling out workspace tree specific queries into a standalone `Query` interface/class, so that it would be easier to mock or fake the dependencies for the provider, like in [not yet rebased one](https://github.com/bazel-contrib/vscode-bazel/pull/422/files#diff-32978083117c650db3d83827ed159c4e5a502efcacd6676a6b5a17f3bc7dd107R48).

If this is too over engineered in this CL I can make this simpler one until it's making more sense in the later changes. WDYT?

src/workspace-tree/bazel_package_tree_item.ts
cwahbong: At first I was going to prevent the type checker and convert the type `IconName | vscode.ThemeIcon` back in the `BazelWorkspaceTreeProvider`, thus moved the logic there. If we want to get the icon path in `BazekTargetTreeItem` then we have to pass the resource class into that as well (maybe it's actually making more sense?).

src/workspace-tree/icons.ts
cwahbong: Do you mean this is not `undefined`? I thought when `iconName` is not found in `SPECIFIC_RULE_CLASS` and either matches in the following suffix rules, then it would be undefined.

vogelsgesang: your annotations are correct. But the TypeScript compiler does not verify them, yet, because we opted out of that check

test/resources.test.ts
cwahbong: At first I was only going to prevent the typo or future inconsistency when we add/remove/adjust the icons.

I still not get how the extension would be installed in `.vscode-test.js`, however when I print out the `extensionPath` it points to the project root. I think we should adjust the test config in #421 to make it test against the `dist/` bundle.

vogelsgesang (approved):

src/workspace-tree/icons.ts
cwahbong: Noted. Let me try enabled it locally while developing new changes.

src/workspace-tree/bazel_package_tree_item.ts
cwahbong: PTAL

src/workspace-tree/bazel_workspace_tree_provider.ts
vogelsgesang: "easier to be tested" - makes sense. Didn't understand the underlying motivation, but now that I understand it, the code is fine as is

vogelsgesang (approved):

Issue #430, opened 2025-01-09
Feature request: URI handling


catskul:
Apparently it's possible to register URI handlers such that something like...

vscode://bazelbuild.vscode-bazel//my/path/to/tests:target

...can be click able from anywhere, and the OS passes that info to vscode and the extension to do with what it likes. In this case simply jumping to the target in the bazel file would be ideal.

https://code.visualstudio.com/api/references/vscode-api#window.registerUriHandler
https://stackoverflow.com/questions/55434556/can-vs-code-extensions-get-uri-requests-from-other-programs

PR #431, opened 2025-01-14
feat: URI Handler

catskul:
Putting my code where my mouth is:

https://github.com/bazel-contrib/vscode-bazel/issues/430

This feature will allow us to add links to documents, and tool output which jumps to specific  targets, reducing friction and wasted time.

URI handler works such that:

[vscode://bazelbuild.vscode-bazel//my/path/to/tests:target](vscode://bazelbuild.vscode-bazel//my/path/to/tests:target)

(unfortunately github itself won't link out to non-http urls)

Will jump to said target given vscode is opened to a workspace which contains a bazel workspace with that target as the first `workspaceFolder`

I imagine we could handle more situations like:

vscode://bazelbuild.vscode-bazel/fspath/to/workspace//my/path/to/tests:target

or

vscode://bazelbuild.vscode-bazel/test//my/path/to/tests:target

But I imagine anything of the sort would be more controversial.

### DIFF
--- a/src/definition/bazel_goto_definition_provider.ts
+++ b/src/definition/bazel_goto_definition_provider.ts
@@ -28,6 +28,37 @@ import { blaze_query } from "../protos";
 // LABEL_REGEX matches label strings, e.g. @r//x/y/z:abc
 const LABEL_REGEX = /"((?:@\w+)?\/\/|(?:.+\/)?[^:]*(?::[^:]+)?)"/;
 
+export async function targetToUri(
+  targetText: string,
+  workingDirectory: Uri,
+): Promise<QueryLocation | undefined> {
+  const match = LABEL_REGEX.exec(targetText);
+
+  const targetName = match[1];
+  // don't try to process visibility targets.
+  if (targetName.startsWith("//visibility")) {
+    return null;
+  }
+
+  const queryResult = await new BazelQuery(
+    getDefaultBazelExecutablePath(),
+    workingDirectory.fsPath,
+  ).queryTargets(`kind(rule, "${targetName}") + kind(file, "${targetName}")`);
+
+  if (!queryResult.target.length) {
+    return null;
+  }
+  const result = queryResult.target[0];
+  let location;
+  if (result.type === blaze_query.Target.Discriminator.RULE) {
+    location = new QueryLocation(result.rule.location);
+  } else {
+    location = new QueryLocation(result.sourceFile.location);
+  }
+
+  return location;
+}
+
 export class BazelGotoDefinitionProvider implements DefinitionProvider {
   public async provideDefinition(
     document: TextDocument,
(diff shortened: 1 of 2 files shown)


src/extension/extension.ts
```diff
@@ -105,6 +105,30 @@ export async function activate(context: vscode.ExtensionContext) {
       "bazel.copyTargetToClipboard",
       bazelCopyTargetToClipboard,
     ),
+    // URI handler
+    vscode.window.registerUriHandler({
+      async handleUri(uri: vscode.Uri) {
+        try {
+          const workspace = vscode.workspace.workspaceFolders[0];
```
catskul, line 112: presume it's in the first `workspaceFolder`. I suppose something more sophisticated could be done, but not sure it's necessary. Happy to accept suggested change here if someone has a specific suggestion.

src/definition/bazel_goto_definition_provider.ts
```diff
@@ -41,33 +74,13 @@ export class BazelGotoDefinitionProvider implements DefinitionProvider {
...
-    if (result.type === blaze_query.Target.Discriminator.RULE) {
-      location = new QueryLocation(result.rule.location);
-    } else {
-      location = new QueryLocation(result.sourceFile.location);
-    }
     return [
       {
         originSelectionRange: range,
-        targetUri: Uri.file(location.path),
+        targetUri: Uri.file(location.path + `:$location.line:$location.column`),
```
catskul, line 83: I think I might not have pushed a change to this which reflects the right way to do it
```suggestion
        targetUri: Uri.file(location.path).with({fragment: `${location.line}:${location.column}`})

```

catskul:
formatting fixes pushed.

cameron-martin:
What's the use case for this?

catskul:
@cameron-martin When targets fail, esp in cloud CI, notifications are often sent via slack, email, etc.

The use case here is to make it possible to provide clickable links right into vscode.

This is directly analogous to VSCode's more general ability to link to files in vscode:

https://github.com/Microsoft/vscode-docs/blob/main/docs/editor/command-line.md#opening-vs-code-with-urls

src/definition/bazel_goto_definition_provider.ts
```diff
@@ -19,15 +19,48 @@ import {
   Position,
   TextDocument,
   Uri,
+  window,
```
cameron-martin, line 22: Both these extra imports seem to be unused.

src/definition/bazel_goto_definition_provider.ts
```diff
@@ -41,33 +74,15 @@ export class BazelGotoDefinitionProvider implements DefinitionProvider {
...
-    // don't try to process visibility targets.
-    if (targetName.startsWith("//visibility")) {
-      return null;
-    }
 
-    const queryResult = await new BazelQuery(
-      getDefaultBazelExecutablePath(),
-      Utils.dirname(document.uri).fsPath,
-    ).queryTargets(`kind(rule, "${targetName}") + kind(file, "${targetName}")`);
+    const location = await targetToUri(targetText, Utils.dirname(document.uri));
```
cameron-martin, line 78: I'm not sure if this case can ever realistically get hit, but to make this refactor preserve behaviour you should return null here if location is null.

src/extension/extension.ts
cameron-martin: Without any way of identifying the bazel workspace, I don't think we can do any better than assume the first workspace, so I'm happy with this.

cameron-martin (changes requested):
Other than those minor comments, look good! Seems to work very well.

catskul:
Will try to update this tonight.

src/definition/bazel_goto_definition_provider.ts
catskul: fixed.

src/definition/bazel_goto_definition_provider.ts
catskul: Fixed with ternary. Let me know if you'd rather I used some other format.

catskul:
All comments addressed. Happy to make any additional changes if you see any other issues.

cameron-martin (approved):

Issue #432, opened 2025-01-28
Starlark debugger is broken in 0.11.0


dpar39:
This change broke the starlark debugger:

https://github.com/bazel-contrib/vscode-bazel/pull/384/files#diff-f3f20d3eedcd2c0127856ac96df3014f449e9e6324f5dcd3b9ae526693f44222R170

How? Firstly `this.bazelInfo` is meant to be a `Map<string, string>>`, not an instance of `BazelInfo`. But the root cause is that the debugger program can't resolve the `vscode` import which got inserted as a dependency from `BazelComand`.

```
node ./out/src/debug-adapter/client.js
node:internal/modules/cjs/loader:1247
  throw err;
  ^

Error: Cannot find module 'vscode'
Require stack:
- /home/dpar39/code/vscode-bazel/out/src/bazel/bazel_command.js
- /home/dpar39/code/vscode-bazel/out/src/bazel/bazel_info.js
- /home/dpar39/code/vscode-bazel/out/src/debug-adapter/client.js
    at Function._resolveFilename (node:internal/modules/cjs/loader:1244:15)
    at Function._load (node:internal/modules/cjs/loader:1070:27)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:217:24)
    at Module.require (node:internal/modules/cjs/loader:1335:12)
    at require (node:internal/modules/helpers:136:16)
    at Object.<anonymous> (/home/dpar39/code/vscode-bazel/out/src/bazel/bazel_command.js:17:16)
    at Module._compile (node:internal/modules/cjs/loader:1562:14)
    at Object..js (node:internal/modules/cjs/loader:1699:10)
    at Module.load (node:internal/modules/cjs/loader:1313:32) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/dpar39/code/vscode-bazel/out/src/bazel/bazel_command.js',
    '/home/dpar39/code/vscode-bazel/out/src/bazel/bazel_info.js',
    '/home/dpar39/code/vscode-bazel/out/src/debug-adapter/client.js'
  ]
}
```

PR #433, opened 2025-01-29
fix: fix starlark debugger broken in release 0.11.0 (#432)

dpar39:
This PR fixes starlark debugging which is broken in release 0.11.0 [#432](https://github.com/bazel-contrib/vscode-bazel/issues/432). The debugger program currently fails to start because it can't import vscode. In addition, it also fixes hovering over symbols with starlark debugger respond with errors because symbols are not defined.

### DIFF
--- a/src/debug-adapter/client.ts
+++ b/src/debug-adapter/client.ts
@@ -15,6 +15,7 @@
 import * as child_process from "child_process";
 import * as fs from "fs";
 import * as path from "path";
+import * as util from "util";
 import {
   ContinuedEvent,
   DebugSession,
@@ -32,7 +33,8 @@ import { DebugProtocol } from "vscode-debugprotocol";
 import { skylark_debugging } from "../protos";
 import { BazelDebugConnection } from "./connection";
 import { Handles } from "./handles";
-import { BazelInfo } from "../bazel/bazel_info";
+
+const execFile = util.promisify(child_process.execFile);
 
 /**
  * Returns a {@code number} equivalent to the given {@code number} or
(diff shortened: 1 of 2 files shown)


src/debug-adapter/client.ts
```diff
@@ -532,6 +537,36 @@ class BazelDebugSession extends DebugSession {
     return bazelExecutable;
   }
 
+  /**
+   * Invokes {@code bazel info} and returns the information in a map.
+   *
+   * @param bazelExecutable The name/path of the Bazel executable.
+   * @param cwd The working directory in which Bazel should be launched.
+   */
+  private async getBazelInfo(
```
cameron-martin, line 546: Why not call `getAll` on the `BazelInfo` object if you want it as a `Map`, instead of re-implementing this?

dpar39: Because `BazelInfo` is a subclass of `BazelCommand` and `BazelCommand` imports the `vscode` API to retrieve settings _but_ this is not available in the regular debugger program, see #432 for the stack trace.

src/debug-adapter/client.ts
```diff
@@ -398,32 +400,35 @@ class BazelDebugSession extends DebugSession {
     args: DebugProtocol.EvaluateArguments,
   ) {
     const threadId = this.frameThreadIds.get(args.frameId);
-
-    const value = (
-      await this.bazelConnection.sendRequest({
+    try {
```
dpar39, line 403: This `try ... catch` is needed to avoid that when hovering over starlark code, vscode automatically requests 'evaluation' of the statement under mouse cursor. Without it, starlark `EvaluateResponse`s such as:
```(22:37:33) DEBUG: Sending debug event:
sequence_number: 33
error {
  message: "name \'target\' is not defined"
}
``` 
causes the debugger program to crash.

On another note, I can't find how to evaluate variables like `target` or `ctx` on the debugger. I see the evaluate request takes the `thread_id` and the `statement`, but I can only successfully evaluate in the global scope of the current thread, i.e., can evaluate call stack local variables. Is this a limitation of the Starlark debugger or something we are missing in the adapter?

src/debug-adapter/client.ts
cameron-martin: It only depends on the settings for the startup options. Maybe we could pass this as a parameter to the constructor instead, to avoid this dependency. Then we could reuse the logic in `BazelInfo`.

cameron-martin: Seems reasonable

cameron-martin (approved):

Issue #435, opened 2025-02-03
Bazelrc Language Server


vogelsgesang:
I started a [language server for bazelrc files](https://github.com/salesforce-misc/bazelrc-lsp) a while ago. By now it is slowly getting into a state where it might be useful for a wider audience.

I am wondering: Should we integrate the bazelrc language server also into this Bazel extension? Or publish it as a separate extension? I would tend towards "separate extension" since this makes it easier to provide new releases, keep bug backlogs separate etc.

But apparently, there was an earlier effort to merge various Bazel extensions into this uber-extension here. I don't know about the motivations of this merge. Can somebody elaborate on those original motivations and provide some guidance on whether to publish `bazelrc-lsp` as a separate extension to the VS-Code market place

laurentlb:
VSCode has a concept of Extension Pack (https://code.visualstudio.com/api/references/extension-manifest#extension-packs). Maybe that's a better fit than bundling everything in one extension?

PR #438, opened 2025-02-15
feat(lsp): Allow passing environment variables to the language server

cameron-martin:
This adds an extra configuration option, `bazel.lsp.env`, which passes extra environment variables to the language server. This is useful to enable logging, for example.

### DIFF
--- a/package.json
+++ b/package.json
@@ -194,6 +194,14 @@
                     "default": [],
                     "description": "The arguments to pass to the LSP executable"
                 },
+                "bazel.lsp.env": {
+                    "type": "object",
+                    "default": {},
+                    "additionalProperties": {
+                        "type": "string"
+                    },
+                    "description": "Environment variables to pass to the LSP executable"
+                },
                 "bazel.trace.server": {
                     "type": "string",
                     "enum": [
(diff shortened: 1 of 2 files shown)


vogelsgesang (approved):

vogelsgesang:
By the way: I have been wondering...
Should we keep everything in a monolithic extension? Or should we distribute it across multiple extensions? E.g. should the language server maybe be a separate VSCode extension?

For the current functionality in this extension (which centers around calling into Bazel into various ways), it makes sense to have it all in one place. Otherwise, code sharing would be pretty hard. But the LSP (both your Starlark language server and my [bazelrc-lsp](https://github.com/bazel-contrib/vscode-bazel/issues/435) rarely invoke the Bazel binary, and don't need to share much code with this extension here.

Also see #435. WDYT? Should we work towards splitting off the language servers from this extension?

cameron-martin:
I'm not too familiar with the trade-offs here. What would be the motivation for splitting this up? 

One thing to consider is that this extension disables some inbuilt functionality if the language server is enabled, on the basis that the language server is likely to provide this. Is this sort of interaction between extensions possible? This wouldn't be an issue if this functionality was removed entirely from the bazel extension and expected to be provided by a language server.

Would each language server be expected to provide it's own extension when there are multiple implementations? If so, this would have the disadvantage of multiple similar extensions, with associated release processes etc, but the advantage of being able to do things specific to the language server, such as shipping the language server with the extension, or auto-downloading it.

Issue #227, opened 2020-12-28
bazel query getting stuck


Michaelhobo:
The Bazel extension occasionally gets stuck. I think it usually happens when bazel crashes once, and leaves the lock file around. I start getting errors that continuously pop up, which look like this:

```
Bazel Build Targets: Command failed: bazel --output_base=/tmp/6a622852f8f9062446271ea7e15aa515 query ...:* --output=package 
Another command holds the client lock: pid=58100 owner=client cwd=/workspace/___
Waiting for it to complete... 
Another command holds the client lock: pid=58342 owner=client cwd=/workspace/___
Waiting for it to complete... 
...
 WARNING: Waiting for server process to terminate (waited 5 seconds, waiting at most 10)
WARNING: Waiting for server process to terminate (waited 10 seconds, waiting at most 10)
INFO: Waited 10 seconds for server process (pid=42461) to terminate.
FATAL: Attempted to kill stale server process (pid=42461) using SIGKILL, but it did not die in a timely fashion.
```

I tried deleting /tmp/616.../lock, but the errors continue showing up.

And I have a lot of bazel processes that keep changing PIDs:

```bash
$ ps aux | grep bazel | wc -l
524
```

For now, I really just need syntax highlighting. Is there a way to disable the continuous queries until this issues is fixed?

sluongng:
Same issue here.

Its also a bit weird that the extension is using a different output_base than what I have in my .bazelrc file

cameron-martin:
@sluongng This extension uses a different output base to prevent concurrent builds from blocking IDE features. See https://bazel.build/run/scripts#output-base-option

mikedanese:
I hit this:
```
$ ps ax | grep bazel |wc -l
     648
```
vscode spanned hundreds of subprocesses all requesting the same query `kind(rule, //kubernetes:all)`.
```
 | | |--- 01964 mike.danese /Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper (Plugin).app/Contents/MacOS/Code Helper (Plugin) /Users/mike.danese/.vscode/
extensions/stylelint.vscode-stylelint-1.3.0/dist/start-server.js --stdio --clientProcessId=1873
 | | |-+- 32934 mike.danese bash bazel query kind(rule, //kubernetes:all) --output=proto --proto:output_rule_attrs='' --noproto:rule_inputs_and_outputs --noproto:default_values
[...]

PR #451, opened 2025-06-07
feat(config): add an option to fully disable queries

zachgrayio:
Adds an option to fully disable queries to support large workspaces where performance is an issue.

In many cases when using this extension in large monorepos, the option to simply disable the functionality that depends on query while keeping basics like syntax highlighting is desireable; this allows the user to avoid accumulating many costly and long running queries that end up consuming 100% of their machine's resource until manually exited, without resorting to uninstalling the plugin.

Related to #227

### DIFF
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -14,7 +14,10 @@
 
 import * as vscode from "vscode";
 
-import { getDefaultBazelExecutablePath } from "../extension/configuration";
+import {
+  getDefaultBazelExecutablePath,
+  areBazelQueriesEnabled,
+} from "../extension/configuration";
 import { IBazelCommandAdapter, IBazelCommandOptions } from "./bazel_command";
 import { BazelQuery } from "./bazel_query";
 import { BazelWorkspaceInfo } from "./bazel_workspace_info";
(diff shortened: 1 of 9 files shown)


src/workspace-tree/bazel_package_tree_item.ts
vogelsgesang, line 1: afaict, this will lead to us showing an empty tree in the target list.

Instead, we should hide the target list completely, if bazel queries are disabled

src/bazel/bazel_quickpick.ts
vogelsgesang, line 1: Bazel queries are used in 3 contexts:
1. In the background, to power codelens and goto definition
2. Upon request, e.g. if a user ran the "Build Target" command
3. When explicitly requested via command variables (configured [as part of the tasks configuration](https://github.com/bazel-contrib/vscode-bazel?tab=readme-ov-file#bazel-tasks))

I see how (1) can trigger a lot of background queries and can lead to unresponsiveness. As such, I do see why we would want to disable them.

For (3), I am pretty sure that we want to keep queries enabled, even if queries for (1) are disabled. After all, the user himself configured those exact queries in his own task configuration.

For (2), I am somewhat on the edge. Commands like "Build Target" need to be explicitly triggered by the user. As such, they don't negatively impact the experience by default. I think we might want to keep the Bazel queries for those enabeld, too

zachgrayio:
I'm catching up after some time off, thanks for the replies, will get back to you shortly in the feedback!

cbandera (changes requested):
@zachgrayio Considering Adrians comment https://github.com/bazel-contrib/vscode-bazel/pull/451#discussion_r2157898756 I think the better way would be to introduce `enableXYZ` config keys for the different features of this plugin to selectively enable or disable those parts that are causing problems or are unneeded.
The current approach renders the functionality broken, which causes a bad (because unclear) user experience

cbandera:
As we are making quite some progress on #490 I will close this PR as "Won't merge". The new extension features will allow you to very selectively disable unneeded functionalities of the Extension.
If you are still missing something that should be possible to deactivate, please comment there.

PR #457, opened 2025-07-10
Enable automatic downloading of buildifier release binaries

### LINKED ISSUE #371: Auto-install tools
Forking this off from a discussion in https://github.com/bazelbuild/vscode-bazel/pull/369#discussion_r1554639007

It would lower the barrier of entry for new users if this extension would offer an option to automatically download any missing tools (bazelisk, buildifier, language server).

lalten:
This allows you to specify something like `v8.2.1` in the `buildifierExecutable` config, and the extension will download the buildifier binary for that release for you.

Fixes https://github.com/bazel-contrib/vscode-bazel/issues/371 (only regarding Buildifier for now, and without any fancy Wasm involvement).

I'm not a Typescript expert so I'm happy to receive feedback.


lalten:
> Have you considered splitting it into three config items?

I did in fact try using option groups and a drop-down for selecting type of value, and value. But that quickly made the code pretty complex.

> And how about defaulting to downloading the latest version if no path/label/version was specified and no executable could be found on the path?

I initially wanted to default to `latest` but the issue is that you don't know if latest has changed, so the extension would need to periodically check if it needs to redownload. Another issue is that the direct asset download links it's constructing now don't work for `latest`.
I guess you'd have to use the GH API to ask what release is latest.
And then again, more complexity, when the current setup is just fine. It's actually kinda nice to not get surprise updates that change formatting I guess.

cbandera:
Thanks for opening and reviving this. Note however that I am currently working on a generalized approach of external tool downloads.

lalten:
The UI selection looks like this now:
grafik
grafik
And settings.json:
grafik
grafik
grafik

lalten:
Ready for another round of review!


> Thanks for opening and reviving this. Note however that I am currently working on a generalized approach of external tool downloads.

@cbandera what's the state of your efforts? Could we land this since it's done already (pending review feedback of course), and then replace with your approach once that's ready?

PR #458, opened 2025-07-16
feat(config): Add an option to limit quick pick queries to the nearest package

geoffreyheller:
Adds an option to limit quick pick queries to the closest package associated with the file currently active in the text editor.

For large projects, this helps improve query performance and relevance by avoiding unnecessary traversal of unrelated packages.   If no file is open, the extension will use the workspace root as the base package.

Set `bazel.limitQuickPickQueriesToNearestPackage` to `true` to enable this feature.

Related to #451


.DS_Store
cbandera, line 1: File added by mistake? Maybe you can directly add it to the `.gitignore` file?

README.md
```diff
@@ -37,6 +37,12 @@ this extension does not automatically _fix_ lint warnings during formatting,
 but you can opt into this by enabling the **Bazel: Buildifier Fix on Format**
 setting.
 
+### Limiting quick pick queries to the nearest package
```
cbandera, line 40: (opinionated) I feel like the separate output base is a configuration that will affect more users then the limiting of quick pick queries. Hence I would place the new config option below the existing one.

src/bazel/bazel_utils.ts
```diff
@@ -144,3 +138,43 @@ export function getBazelWorkspaceFolder(fsPath: string): string | undefined {
...
+ * the given file.
+ *
+ * @param fsPath The path to a file in a Bazel workspace.
+ * @returns The Bazel path to the directory with the BUILD.Bazel if found,
+ * others undefined.
+ */
+export function getBazelPackageFolder(fsPath: string): string | undefined {
+  // These are the names of the files that mark the root of a repository
+  // or workspace.
+  const PACKAGE_FILE_NAMES = ["BUILD.bazel"];
```
cbandera, line 177: Both variants are valid:
```suggestion
  const PACKAGE_FILE_NAMES = ["BUILD.bazel", "BUILD"];
```

src/bazel/bazel_utils.ts
```diff
@@ -144,3 +138,43 @@ export function getBazelWorkspaceFolder(fsPath: string): string | undefined {
 
   return undefined;
 }
+
+/**
+ * Search for the path to the directory that has the Bazel WORKSPACE file for
+ * the given file.
```
cbandera, line 144:
```suggestion
 * Search for the path to the directory that has a file marking the root of a Bazel workspace
```

src/bazel/bazel_utils.ts
```diff
@@ -144,3 +138,43 @@ export function getBazelWorkspaceFolder(fsPath: string): string | undefined {
...
+ *
+ * If multiple directories along the path to the file have files called
+ * "WORKSPACE", the lowest path is returned.
```
cbandera, line 147:
```suggestion
 * If multiple directories along the path to the file have matching files, the closest one is returned.
```

src/bazel/bazel_utils.ts
```diff
@@ -94,17 +94,21 @@ function shouldIgnorePath(fsPath: string): boolean {
...
- * @param fsPath The path to a file in a Bazel workspace.
- * @returns The path to the directory with the Bazel WORKSPACE file if found,
- * others undefined.
+ * @param fsPath The path to a file to use as a starting point for the search.
+ * @param targetFiles The names of the files to look for
+ * @returns The path to the directory where the target file is found, or
+ * undefined if the target file is not found.
  */
-export function getBazelWorkspaceFolder(fsPath: string): string | undefined {
+function getTargetFolderForFile(
```
cbandera, line 108: The word `target` is already overloaded in the Bazel ecosystem. How about `findAncestorDirectoryContaining` or something similar that better reflects the intention?

package.json
```diff
@@ -149,6 +149,11 @@
                     "default": false,
                     "markdownDescription": "Whether to add a CodeLens to `BUILD`/`BUILD.bazel` files to provide actions while browsing the file."
                 },
+                "bazel.limitQuickPickQueriesToNearestPackage": {
+                    "type": "boolean",
+                    "default": false,
+                    "markdownDescription": "Limit quick pick queries to the nearest package."
```
cbandera, line 155: Let's extend the description here a little bit, so that it is not the exact same as the heading. How about:
```suggestion
                    "markdownDescription": "If active, the selection dialog for targets or packages will limit the results to the current package."
```

src/bazel/bazel_quickpick.ts
```diff
@@ -146,8 +166,8 @@ export async function queryQuickPickPackage({
 
   const packagePaths = await new BazelQuery(
     getDefaultBazelExecutablePath(),
-    workspaceInfo.workspaceFolder.uri.fsPath,
-  ).queryPackages(query ?? "//...");
+    getWorkingDirectory(workspaceInfo),
```
cbandera, line 169: Is it actually expected to limit the results to just the current package for this `queryQuickPickPackage` function? This means that the result will always be of length 1

cbandera (commented):
Thank you for your contribution! I have left some comments in the code, some are simply my opinion and you can of course decide on whether you agree.

As for the feature itself: I feel like limiting to the current package only could be a little to strict for many users and that the mentioned project file would be the better way to go. Maybe: Would it already serve your usecase if the list of results is ordered in a way that shows the targets of the current package first and the rest later?

Note: I am not a maintainer, so consider this just as thoughts for discussion

Issue #383, opened 2024-04-17
Extension adds MODULE.bazel and MODULE.bazel in local repositories


smocherla-brex:
Maybe this is expected behavior but I wanted to still check and confirm. Anytime I open a bzl file in a workspace within the repo which is a `local_repository`, it ends up creating the MODULE.bazel and MODULE.bazel.lock. We haven't migrated to bzlmod and have `--noenable_bzlmod` in the root workspace but these local repositories don't have it and there's no `.bazelversion` file, so it ends up adding these 2 files every single time and we have to keep undoing it. Is this expected behavior and we need to setup `.bazelrc` and `.bazelversion` in each of our local repositories? Or is it unintended and actually a bug?

cameron-martin:
There's a slack thread about this here: https://bazelbuild.slack.com/archives/C014RARENH0/p1702897315236089

It's a tricky problem to solve. Out of interest, what situation do you have? Is the `local_repository` valid on it's own or should it only be used from the root workspace?

smocherla-brex:
Interesting, thanks for that link. In general, nearly all of our usecases for `local_repository` are not self-contained i.e they're solely for the purpose of the main repository/workspace and they're rarely/never built on their own, so it seemed a like surprising change in workflow for us. We do have 1 `local_repository` in our repo which is valid on its own because we mirror changes downstream as a ruleset for others to use.

I wonder if we could add these workspace repos to `.bazelignore` at the root of the main workspace and the extension could respect that? That said, it seems like it's complex/tricky as you said due to a variety of usecases that need to be taken into account, so we could probably just add `.bazelrc` files in all these local repos.

PR #461, opened 2025-07-31
feat(workspace-tree): Sync open editor with workspace tree view

cbandera:
This PR implements a synchronization between the file opened in the editor and the workspace tree view of the bazel extension.

Note on implementation:
The implementation of reveal() in the tree view required proper parent-child relationships and package path resolution. The changes ensure that:

1. The tree view can properly navigate and reveal items by maintaining correct parent references
2. Package paths are consistently resolved through the parent chain
3. The selected item in the tree view is synchronized with the currently open editor

A cache of BazelPackageTreeItems was added to improve performance by:

- Reducing redundant package path calculations
- Avoiding bazel queries performed by existing getChildren operator
- Minimizing repeated string operations when resolving package paths

This PR is part of #353


cbandera:
@cameron-martin , @vogelsgesang I assumed the tests might be failing because the github runner env doesn't have `bazel` installed, but adding it to the workflow file didn't solve the problem. The tests run through locally (even though I got the same error once). So they might also be flaky - I have the feeling it could have something to do with timing. Maybe we need to wait longer inside the testcase, but I am unsure how and where to do it. 
Any hints on how to debug this are welcome.

cbandera:
@cameron-martin , @vogelsgesang I have deactivated the one problematic test case for now, so that this PR becomes green anyway. 
As stated before, the feature works when manually starting the extension and performing this test case, I just can't get the automated test to work, probably due to timing/concurrency issues.
If you have the time to analyze or even fix the testcase, feel free to do so. Otherwise I am also happy for any advice you can give.

cameron-martin:
Looks like the tests are failing even though you disabled the problematic test. Could you take a look at this?

cbandera:
Ok found it and fixed it. The problem was that due to the way the test was setup, there were multiple instances of the treeview provider which messed things up.
I now changed it to use the one single instance of the running extension for testing. Which fixes all tests.

src/workspace-tree/bazel_package_tree_item.ts
```diff
@@ -67,21 +66,27 @@ export class BazelPackageTreeItem
       return new BazelTargetTreeItem(
         this.resources,
         this.workspaceInfo,
+        this as unknown as IBazelTreeItem,
```
cameron-martin: I don't think we need these type-casts, because this already implements `IBazelTreeItem`.

src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -191,10 +227,37 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
       return new BazelTargetTreeItem(
         this.resources,
         this.workspaceInfo,
+        this as unknown as IBazelTreeItem,
```
cameron-martin: This one here isn't needed either.

src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -55,17 +63,36 @@ export class BazelWorkspaceTreeProvider
       false,
       false,
     );
+
+    // Initialize disposables array if it's not already initialized
```
cameron-martin: What's the point of initialising it like this, rather than like how it used to be?

src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -158,8 +185,103 @@ export class BazelWorkspaceTreeProvider
...
+      (item) =>
+        item.getWorkspaceInfo().workspaceFolder.uri.toString() ===
+        workspaceFolderVSCode.uri.toString(),
+    );
+    if (!workspaceFolderTreeItem) {
+      return undefined; // File does not belong to a detected bazel workspace
+    }
+
+    const relativeFilePath = path.relative(
+      workspaceFolderVSCode.uri.fsPath,
```
cameron-martin, line 236: Should this be relative to the vscode workspace, or the bazel workspace?

src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -158,8 +185,103 @@ export class BazelWorkspaceTreeProvider
...
+
+    if (this.lastSelectedUri?.toString() === fileUri.toString()) {
+      return; // Already processed this file
+    }
+
+    try {
+      const packageItem = this.getPackageTreeItemFromUri(fileUri);
+      if (packageItem) {
+        this.lastSelectedUri = fileUri;
+        await this.revealTreeItem(packageItem as IBazelTreeItem);
```
cameron-martin: This cast isn't necessary either.

src/workspace-tree/bazel_workspace_tree_provider.ts
cbandera: None. I reverted it. Was a leftover from my endless tries to get the tests running again.

src/workspace-tree/bazel_workspace_tree_provider.ts
cbandera: Thanks. I removed all unnecessary casts

src/workspace-tree/bazel_workspace_tree_provider.ts
cbandera: At the moment relative to the vscode workspace.
Is it a requirement to work with bazel workspaces that do not live at the root of the vscode workspace?
If yes, then the current implementation is already not working...

But I have tested the usecase when several Folders are open in a workspace (each with a bazel workspace and tragets). This usecase is supported as I understood from the function `updateWorkspaceFolderTreeItems` and the synchronization for this usecase with 2 bazel trees works fine.

cameron-martin: That's a good question. It'd be nice if we didn't have to assume a 1:1 mapping, but yeah I imagine we're quite inconsistent/buggy in this respect. For example sometimes we run bazel commands in the workspace that the current file resides in, instead of the vscode workspace root, causing issues like #383.

Happy to defer this to later though. From prior discussions I've had doing workspace discovery "correctly" is hard.

cameron-martin (approved):

PR #463, opened 2025-07-31
feat(navigation): add "Jump To Label" command

cbandera:
This PR adds a new command `Jump To Label`, that allows users to quickly jump to the target declaration for a given label.

To implement this, the result of a `BazelQuery` is extended with the proto message that was returned from the subprocess call.

Part of #353.

BEGIN_COMMIT_OVERRIDE
feat(navigation): Add "Go to Label" command
END_COMMIT_OVERRIDE


cameron-martin (approved):

cameron-martin:
Okay no problem, I'm happy to open a PR.

Issue #300, opened 2023-07-21
[Feature request] Add option to filter CodeLens links in BUILD files


rossc719g:
TL;DR: It would be really nice to have a way to filter the displayed CodeLens links.

In BUILD files, there are often extra hidden rules defined by rules.

For example, if you use:
```
load("@bazel_skylib//rules:build_test.bzl", "build_test")

build_test(
  name="foo"
  ...
)
```
You get the expected `Test //whatever:foo` link, but you also get `Build //whatever:foo_0__deps`.

And if you have functions in a `.bzl` file that create multiple sub-rules, the list of links can quickly get very long, and end up wider than the available space, and the UI does not let you scroll right if your file content is not that wide.
(Incidentally, it would be nice to wrap these long lines of links, but that is a separate FR)

Ideally, it would just honor the existing `Bazel › Command Line: Query Expression` option for these links as well.
But it would be fine to have a similar option specifically for the CodeLens links if that is deemed to be nicer.

Thoughts?

KrauseStefan:
This would be really nice to have both for the CodeLens links but also for the side bar tree.

The Idea plugin has this trough the `no-ide` tag.
https://ij.bazel.build/docs/tags.html#no-ide

Something similar could be very nice.
Alternatively just a setting to add filter by name.

What something like this be a welcome change for the maintainers?

vogelsgesang:
> Alternatively just a setting to add filter by name.

For the side bar / the "Bazel Build Targets" tree, there is already the setting `bazel.commandLine.queryExpression`.
I guess you could write a Bazel query to filter out `no-ide` tagged targets?

theswit:
Perhaps simpler, can we put the copy/build/run/test option of the `name` target first? I'm hoping if we can't get proper filtering support, maybe just adding a sort change could gain more traction? query expressions may be pretty difficult to get right across a large or complex build codebase. but I suspect most folks care most about the named target itself; I know that's my primary use case.

cbandera:
I think a great solution for this problem was found and implemented with #477 . If you still see the need to filter the results further please comment/reopen the issue.

theswit:
+1, I just noticed this the other day. It was a bit jarring at first, and I wasn't sure if it was related to a new vscode release (1.107) or an extension update. But after the initial 'what the', it's fantastic!

PR #464, opened 2025-07-31
feat(codelens): Sort Bazel build targets by name length in codelens

cbandera:
This change modifies the BazelBuildCodeLensProvider to sort targets by the
length of their names (shortest first) before displaying them in the codelens
menu.

The motivation is to provide a more predictable and user-friendly experience
when working with macros in BUILD files. When a macro generates multiple
targets, the main target (which typically has the shortest name) will now
consistently appear first in the codelens menu, while helper targets
(which usually have longer names with suffixes) will appear afterward.

This is particularly useful in scenarios where a macro generates both a main
target and several internal helper targets, as it ensures the most relevant
target is always presented first to the user.

Part of #300

### DIFF
--- a/src/codelens/bazel_build_code_lens_provider.ts
+++ b/src/codelens/bazel_build_code_lens_provider.ts
@@ -137,7 +137,16 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
         countMap.set(line, countMap.has(line));
         return countMap;
       }, new Map<number, boolean>());
-    for (const target of queryResult.target) {
+    // Sort targets by length first, then alphabetically
+    // This ensures shorter names (often main targets) appear first, with consistent ordering within each length group
+    const sortedTargets = [...queryResult.target].sort((a, b) => {
+      const lengthDiff = a.rule.name.length - b.rule.name.length;
+      return lengthDiff !== 0
+        ? lengthDiff
+        : a.rule.name.localeCompare(b.rule.name);
+    });
+
+    for (const target of sortedTargets) {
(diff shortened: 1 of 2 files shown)


src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -137,7 +137,13 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
         countMap.set(line, countMap.has(line));
         return countMap;
       }, new Map<number, boolean>());
-    for (const target of queryResult.target) {
+    // Sort targets by name length
+    // This is a heuristic to achieve that the main target generated by marcros will also be the first shown by the codelens
+    const sortedTargets = [...queryResult.target].sort(
+      (a, b) => a.rule.name.length - b.rule.name.length,
```
cameron-martin: How about instead sorting by `a.rule.name.localeCompare(b.rule.name)`? This way, they're sorted alphabetically but also retain the property where if `a.rule.name` is a prefix of `b.rule.name`, then `a` comes first. I think this is the main case you're talking about, where a macro generates a target `foo`, but also `foo_bar` and `foo_baz`.

cbandera: I have added the localCompare in addition to the sorting by lenght. The reasoning is that I am assuming postfixes, but someone could also add prefixes in a macro. In this case the alphabetical sorting might have the contrary effect.

So now, the targets are sorted by length first and then alphabetically.

I also added tests to demonstrate the behavior.

cameron-martin (approved):

PR #465, opened 2025-07-31
feat(quickpick): Introduce dynamic quickpick for targets/packages.

cbandera:
Dynamic quickpick will run bazel queries in the background as the user types.

This will limit the query for relevant labels to just a subset of the repository, which can greatly improve feedback time in case of huge bazel workspaces. 

The default behaviour is to start with querying the complete workspace (as before) but to cancel a running query as soon as the user starts typing and to start a new query with the given input.

Part of #353. To be merged after #463

### DIFF
--- a/src/extension/bazel_wrapper_commands.ts
+++ b/src/extension/bazel_wrapper_commands.ts
@@ -13,11 +13,16 @@
 // limitations under the License.
 
 import * as vscode from "vscode";
+import * as path from "path";
 
 import { IBazelCommandAdapter } from "../bazel/bazel_command";
 import { BazelWorkspaceInfo } from "../bazel/bazel_workspace_info";
 import { getDefaultBazelExecutablePath } from "./configuration";
-import { getBazelPackageFile } from "../bazel/bazel_utils";
+import {
+  getBazelPackageFile,
+  getBazelWorkspaceFolder,
+  getBazelPackageFolder,
+} from "../bazel/bazel_utils";
 import {
   queryQuickPickTargets,
   queryQuickPickPackage,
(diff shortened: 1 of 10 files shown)


src/bazel/bazel_query.ts
```diff
@@ -143,27 +149,83 @@ export class BazelQuery extends BazelCommand {
...
+        }
+      };
+
+      if (abortSignal) {
+        if (abortSignal.aborted) {
+          onAbort();
+          return;
+        }
+        // Type assertion to access the non-standard addEventListener
+        const signal = abortSignal as unknown as {
```
cameron-martin: If you upgrade `@types/node` to latest you shouldn't have to do these type casts.

src/bazel/bazel_quickpick.ts
```diff
@@ -169,10 +173,108 @@ export async function queryQuickPickPackage({
...
+
+    try {
+      const items = await options.queryFunctor({
+        query: newQuery,
+        workspaceInfo: options.workspaceInfo,
+        abortSignal: currentAbortController.signal,
+      });
+      quickPick.items = items;
+    } catch (error) {
+      if (error instanceof Error && error.message === "Query was aborted") {
```
cameron-martin: Checking the message here seems slightly brittle. If we instead rejected with the AbortSignal `reason` (which by default is a AbortError DOMException), then we could check for this, e.g. by using `error.name == "AbortError"`.

src/bazel/bazel_quickpick.ts
```diff
@@ -169,10 +173,108 @@ export async function queryQuickPickPackage({
...
+    const currentAbortController = new AbortController();
+    abortController = currentAbortController;
+
+    // Show loading state
+    quickPick.busy = true;
+    quickPick.items = [];
+
+    // Process the input: keep everything before the last slash and add '...'
+    const newQuery = value.includes("/")
+      ? options.query.replace(
```
cameron-martin: Looks like we sometimes don't pass a query here but don't account for this by defaulting `options.query`. Note we don't have strict null checks enabled so this won't be caught by typescript unfortunately!

src/extension/bazel_wrapper_commands.ts
```diff
@@ -36,12 +37,11 @@ async function bazelBuildTarget(adapter: IBazelCommandAdapter | undefined) {
...
     // invoked via the command palatte. Provide quickpick build targets for
     // the user to choose from.
-    const quickPick = await vscode.window.showQuickPick(
-      queryQuickPickTargets({ query: "kind('.* rule', ...)" }),
-      {
-        canPickMany: false,
-      },
-    );
+    const quickPick = await showDynamicQuickPick({
+      query: "kind('.* rule', //...)",
```
cameron-martin: To me using `//...` as a placeholder here isn't that obvious. How about if this were instead a function, for example:

```
pattern => `kind('.* rule', ${pattern})`
```

cameron-martin (changes requested):

src/bazel/bazel_query.ts
cbandera: Thanks I did so and it makes the code easier to read. However I also had to pull up the version of ESLint so that it would no longer complain about the type either. This in turn made it necessary to update the eslint config and one other type hint in the source code... 
Let me know whether you want to keep these changes or whether I should only pull up `node` and suppress the eslint warnings instead.

src/bazel/bazel_quickpick.ts
cbandera: Adressed.

src/bazel/bazel_quickpick.ts
cbandera: Good catch! Adressed

src/extension/bazel_wrapper_commands.ts
cbandera: Great idea. I changed the interface to take both: A pattern and a builder function. I think this makes the code easier to understand.

PR #468, opened 2025-08-05
fix: Use "go to" instead of "jump"

cameron-martin:
This changes the terminology used in various commands from "go to" to "jump". This terminology is already used all over vscode, whereas jump is not. This was agreed in #463.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,13 +88,13 @@
             },
             {
                 "category": "Bazel",
-                "command": "bazel.jumpToBuildFile",
-                "title": "Jump to BUILD file"
+                "command": "bazel.goToBuildFile",
+                "title": "Go to BUILD file"
             },
             {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -88,13 +88,13 @@
...
-                "title": "Jump to BUILD file"
+                "command": "bazel.goToBuildFile",
+                "title": "Go to BUILD file"
             },
             {
                 "category": "Bazel",
-                "command": "bazel.jumpToLabel",
-                "title": "Jump to Bazel Label"
+                "command": "bazel.goToLabel",
+                "title": "Go to Label"
```
cameron-martin, line 97: The "Bazel" here wasn't necessary because commands are already prefixed with their category, so it used to redundantly show:

> Bazel: Jump to Bazel Label

PR #477, opened 2025-09-22
feat(codelens): Show picker for multiple targets

bkolb:
This change modifies the BazelBuildCodeLensProvider to create a target picker in case there is more than one target per category (Copy/Build/Run/Test).

The motivation is to provide a better and more user-friendly experience when working with macros in BUILD files. When a macro generates multiple targets, you could run into situations where not all actions were visible anymore.

### DIFF
--- a/src/codelens/bazel_build_code_lens_provider.ts
+++ b/src/codelens/bazel_build_code_lens_provider.ts
@@ -20,26 +20,24 @@ import { getDefaultBazelExecutablePath } from "../extension/configuration";
 import { blaze_query } from "../protos";
 import { CodeLensCommandAdapter } from "./code_lens_command_adapter";
 
-/** Computes the shortened name of a Bazel target.
+/**
+ * Groups of Bazel targets organized by the actions they support.
+ * Used by the CodeLens provider to determine which actions to display for each target.
  *
- * For example, if the target name starts with `//foo/bar/baz:fizbuzz`,
- * the target's short name will be `fizzbuzz`.
- *
- * This allows our code lens suggestions to avoid filling users' screen with
- * redundant path information.
- *
- * @param targetName The unshortened name of the target.
- * @returns The shortened name of the target.
+ * @interface ActionGroups
+ * @property {string[]} copy - Targets that support copying their label to clipboard (all target types)
+ * @property {string[]} build - Targets that support build operations (libraries, binaries, tests)
+ * @property {string[]} test - Targets that support test execution (test rules only)
+ * @property {string[]} run - Targets that support run operations (executable binaries only)
  */
-function getTargetShortName(targetName: string): string {
-  const colonFragments = targetName.split(":");
-  if (colonFragments.length !== 2) {
-    return targetName;
-  }
-  return colonFragments[1];
+interface ActionGroups {
+  copy: string[];
+  build: string[];
+  test: string[];
+  run: string[];
 }
 
-/** Provids CodeLenses for targets in Bazel BUILD files. */
+/** Provides CodeLenses for targets in Bazel BUILD files. */
 export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
   public onDidChangeCodeLenses: vscode.Event<void>;
 
@@ -117,8 +115,9 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
    * Takes the result of a Bazel query for targets defined in a package and
    * returns a list of CodeLens for the BUILD file in that package.
    *
-   * @param bazelWorkspaceDirectory The Bazel workspace directory.
-   * @param queryResult The result of the bazel query.
+   * @param bazelWorkspaceInfo The Bazel workspace information containing workspace path and context
+   * @param queryResult The result of the bazel query containing target definitions
+   * @returns A new array of CodeLens objects for the BUILD file
    */
   private computeCodeLenses(
     bazelWorkspaceInfo: BazelWorkspaceInfo,
@@ -126,50 +125,108 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
   ): vscode.CodeLens[] {
     const result: vscode.CodeLens[] = [];
 
-    interface LensCommand {
-      commandString: string;
-      name: string;
-    }
-
-    const useTargetMap = queryResult.target
-      .map((t) => new QueryLocation(t.rule.location).line)
-      .reduce((countMap, line) => {
-        countMap.set(line, countMap.has(line));
-        return countMap;
-      }, new Map<number, boolean>());
-    // Sort targets by length first, then alphabetically
-    // This ensures shorter names (often main targets) appear first, with consistent ordering within each length group
+    // Sort targets alphabetically
     const sortedTargets = [...queryResult.target].sort((a, b) => {
-      const lengthDiff = a.rule.name.length - b.rule.name.length;
-      return lengthDiff !== 0
-        ? lengthDiff
-        : a.rule.name.localeCompare(b.rule.name);
+      return a.rule.name.localeCompare(b.rule.name);
     });
 
+    // Group targets by line number to handle multiple targets on same line
+    const targetsByLine = new Map<number, typeof sortedTargets>();
     for (const target of sortedTargets) {
       const location = new QueryLocation(target.rule.location);
+      const line = location.line;
+      if (!targetsByLine.has(line)) {
+        targetsByLine.set(line, []);
+      }
+      targetsByLine.get(line)?.push(target);
+    }
+
+    // Process each line's targets
+    for (const [, targets] of targetsByLine) {
+      this.createCodeLensesForTargetsOnSameLine(
+        targets,
+        bazelWorkspaceInfo,
+        result,
+      );
+    }
+
+    return result;
+  }
+
+  /**
+   * Creates CodeLens objects for targets on the same line.
+   *
+   * @param targets Array of Bazel targets found on the same line in the BUILD file
+   * @param bazelWorkspaceInfo Workspace context information for command creation
+   * @param result Output array that will be modified in-place to include new CodeLens objects
+   */
+  private createCodeLensesForTargetsOnSameLine(
+    targets: blaze_query.ITarget[],
+    bazelWorkspaceInfo: BazelWorkspaceInfo,
+    result: vscode.CodeLens[],
+  ): void {
+    const location = new QueryLocation(targets[0].rule.location);
+
+    const actionGroups = this.groupTargetsByAction(targets);
+
+    this.createCodeLens(
+      "Copy",
+      "bazel.copyLabelToClipboard",
+      actionGroups.copy,
+      location,
+      bazelWorkspaceInfo,
+      result,
+    );
+    this.createCodeLens(
+      "Build",
+      "bazel.buildTarget",
+      actionGroups.build,
+      location,
+      bazelWorkspaceInfo,
+      result,
+    );
+    this.createCodeLens(
+      "Test",
+      "bazel.testTarget",
+      actionGroups.test,
+      location,
+      bazelWorkspaceInfo,
+      result,
+    );
+    this.createCodeLens(
+      "Run",
+      "bazel.runTarget",
+      actionGroups.run,
+      location,
+      bazelWorkspaceInfo,
+      result,
+    );
+  }
+
+  /**
+   * Groups targets by the actions they support based on Bazel rule types.
+   * Uses rule naming conventions to determine which actions are available.
+   *
+   * @param targets Array of Bazel targets to classify by supported actions
+   * @returns ActionGroups object with targets organized by action type
+   */
+  private groupTargetsByAction(targets: blaze_query.ITarget[]): ActionGroups {
+    const copyTargets: string[] = [];
+    const buildTargets: string[] = [];
+    const testTargets: string[] = [];
+    const runTargets: string[] = [];
+
+    for (const target of targets) {
       const targetName = target.rule.name;
       const ruleClass = target.rule.ruleClass;
-      const targetShortName = getTargetShortName(targetName);
 
-      const commands: LensCommand[] = [];
-
-      // All targets support target copying and building.
-      commands.push({
-        commandString: "bazel.copyLabelToClipboard",
-        name: "Copy",
-      });
-      commands.push({
-        commandString: "bazel.buildTarget",
-        name: "Build",
-      });
+      // All targets support copying and building
+      copyTargets.push(targetName);
+      buildTargets.push(targetName);
 
       // Only test targets support testing.
       if (ruleClass.endsWith("_test") || ruleClass === "test_suite") {
-        commands.push({
-          commandString: "bazel.testTarget",
-          name: "Test",
-        });
+        testTargets.push(targetName);
       }
 
       // Targets which are not libraries may support running.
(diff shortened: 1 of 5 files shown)


src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -146,62 +141,200 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
         : a.rule.name.localeCompare(b.rule.name);
```
cbandera: I guess it is to be discussed again, whether we want to keep this "workaround" that I introduced with the goal of putting the most relevant target first in the list of codelenses.
With the proposed quick pick menu, I feel like a purely alphabetical ordering is much more intuitive.

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -146,62 +141,200 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
+      const hasMultipleTargets = useTargetMap.get(line);
+
+      if (hasMultipleTargets) {
+        // For multiple targets on same line, create grouped CodeLens with dropdowns
+        this.createGroupedCodeLenses(targets, bazelWorkspaceInfo, result);
+      } else {
+        // For single target on line, use existing logic
+        const target = targets[0];
+        this.createSingleTargetCodeLenses(target, bazelWorkspaceInfo, result);
+      }
```
cbandera: I would prefer to merge these two methods to reduce code duplication. I believe we can treat the two code paths mostly the same and try to "push" the if-else construct to a lower helper function.

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -146,62 +141,200 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
       const location = new QueryLocation(target.rule.location);
-      const targetName = target.rule.name;
-      const ruleClass = target.rule.ruleClass;
-      const targetShortName = getTargetShortName(targetName);
+      const line = location.line;
+      if (!targetsByLine.has(line)) {
+        targetsByLine.set(line, []);
+      }
+      targetsByLine.get(line)?.push(target);
+    }
```
cbandera, line 142: Can this be merged with, or replace the computation of `useTargetMap`? It is essentially doing the same thing. And in line 157 the two variables are even mixed up in usage

src/codelens/bazel_build_code_lens_provider.ts
cbandera: In fact I believe you can already simply delete the "single" branch, as the "grouped" branch is also capable of handling arrays of size 1 (except the tooltip)

src/extension/bazel_wrapper_commands.ts
```diff
@@ -34,9 +37,55 @@ import { blaze_query } from "../protos";
  * Builds a Bazel target and streams output to the terminal.
  *
  * @param adapter An object that implements {@link IBazelCommandAdapter} from
- * which the command's arguments will be determined.
+ * which the command's arguments will be determined, or BazelWorkspaceInfo
+ * for the new unified format from code lens calls.
```
cbandera: It is not clear to me, why you had to move away from the existing `IBazelCommandAdapter` interface, which was used via the `CodeLensCommandAdapter`. From what I can tell, the existing Adapter is capable of achieving the same, without modification.
By keeping the old interface, a lot of the complexity in this file could be reduced.

src/extension/bazel_wrapper_commands.ts
```diff
@@ -34,9 +37,55 @@ import { blaze_query } from "../protos";
...
+    if (targetsInfo.targets.length === 1) {
+      // Single target - build directly
+      const buildCommandOptions: IBazelCommandOptions = {
+        targets: targetsInfo.targets,
+        options: [],
+        workspaceInfo: bazelWorkspaceInfo,
+      };
+      const buildTask = createBazelTask("build", buildCommandOptions);
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      vscode.tasks.executeTask(buildTask);
```
cbandera: This block is repeated further below. Let's avoid this code duplication here (and in the other functions below) by turning the order around and collect the single target to be build first (either directly or via quickpick) and afterwards continue to buildup the task and execute it.

src/extension/bazel_wrapper_commands.ts
```diff
@@ -447,6 +627,92 @@ function bazelCopyLabelToClipboard(adapter: IBazelCommandAdapter | undefined) {
...
+      }),
+    };
+    await commandFunction(adapter);
+  }
+}
+
+/**
+ * Multiple target copy command
+ */
+async function bazelCopyLabelToClipboardMultiple(
```
cbandera: This function seems unused

src/extension/bazel_wrapper_commands.ts
```diff
@@ -447,6 +627,92 @@ function bazelCopyLabelToClipboard(adapter: IBazelCommandAdapter | undefined) {
...
+    label: formatTargetDisplayName(target),
+    description: target, // Full target path as description
+    target,
+  }));
+}
+
+/**
+ * Shows a QuickPick for multiple targets and executes the selected command
+ */
+async function bazelMultipleTargetCommand(
```
cbandera: This function already bundles the logic of differentiating between multiple labels. Might this be a candidate for integrating with the `CodeLensCommandAdapter`? I believe if you abstract the problem of selecting one of multiple away like this, you could leave almost all code in this file as is.

src/extension/bazel_wrapper_commands.ts
```diff
@@ -380,12 +524,48 @@ async function bazelGoToLabel(target_info?: blaze_query.ITarget | undefined) {
...
+      typeof targetsOrUndefined === "string"
+        ? [targetsOrUndefined]
+        : (targetsOrUndefined as { targets: string[] }).targets;
+
+    if (targets.length === 1) {
+      // Single target - copy directly
+      label = targets[0];
+    } else {
+      // Multiple targets - show selection dialog
+      return bazelMultipleTargetCommand(
```
cbandera: Why use a separated helper function here, but not in the other functions? (build, run, test)

src/codelens/bazel_build_code_lens_provider.ts
bkolb: Replaced length-based + alphabetical sorting with pure alphabetical sorting

src/codelens/bazel_build_code_lens_provider.ts
bkolb: I created `createCodeLensesForTargets` with a shared helper function `groupTargetsByAction`

src/codelens/bazel_build_code_lens_provider.ts
bkolb: Done

src/extension/bazel_wrapper_commands.ts
bkolb: I restored `IBazelCommandAdapter`

src/extension/bazel_wrapper_commands.ts
bkolb: I extracted `executeBazelCommand` helper function

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -126,50 +124,73 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
+
+    // Process each line's targets
+    for (const [, targets] of targetsByLine) {
+      // Use unified method for both single and multiple targets
+      this.createCodeLensesForTargets(targets, bazelWorkspaceInfo, result);
+    }
+
+    return result;
```
cbandera, line 153: Out of curiosity: I would have probably returned from `createCodeLensesForTargets` and assigned to `result` instead of making `result` a `in-out` argument. While both approaches work, is there a benefit to doing it this way?

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -126,50 +124,73 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
+      if (!targetsByLine.has(line)) {
+        targetsByLine.set(line, []);
+      }
+      targetsByLine.get(line)?.push(target);
+    }
+
+    // Process each line's targets
+    for (const [, targets] of targetsByLine) {
+      // Use unified method for both single and multiple targets
+      this.createCodeLensesForTargets(targets, bazelWorkspaceInfo, result);
```
cbandera: Minor: How about naming this function `createCodeLensesForTargetsOnSameLine`

src/codelens/bazel_build_code_lens_provider.ts
cbandera, line 122: Minor: How about naming this function `createCodeLensesForDocument`

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -126,50 +124,73 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
+    bazelWorkspaceInfo: BazelWorkspaceInfo,
+    result: vscode.CodeLens[],
+  ): void {
+    const location = new QueryLocation(targets[0].rule.location);
+
+    // Group targets by action type using shared logic
+    const actionGroups = this.groupTargetsByAction(targets);
+
+    // Use unified method for all targets (handles both single and multiple)
+    this.createMultipleTargetCodeLenses(
```
cbandera: Minor: ...because I am having a hard time coming up with a good name: Maybe its better to avoid another helper function here and just directly paste the 4 statements directly in this function?

src/codelens/bazel_build_code_lens_provider.ts
```diff
@@ -180,28 +201,86 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
...
+      location,
+      bazelWorkspaceInfo,
+      result,
+    );
+  }
+
+  /**
+   * Creates a CodeLens for a specific action type if targets are available.
+   */
+  private createActionCodeLens(
```
cbandera: Minor: How about naming this function simply `createCodeLens`

src/extension/bazel_wrapper_commands.ts
```diff
@@ -29,37 +29,72 @@ import {
...
   vscode.tasks.executeTask(task);
 }
 
+/**
+ * Builds a Bazel target and streams output to the terminal.
+ *
+ * @param adapter An object that implements {@link IBazelCommandAdapter} from
+ * which the command's arguments will be determined.
+ */
+async function bazelBuildTarget(adapter: IBazelCommandAdapter | undefined) {
```
cbandera, line 93: I am still having a hard time following the logic of this code and I have the feeling that functionality is still duplicated or obfuscated.
My understanding is that we have three different usecases that we want to support:
- Function is invoked through command palette without a specified target: 
  - Adapter is undefined and we want to prompt the user for a target from the scope specified by the query
- Function is invoked through some internal command (e.g. a codelens) with a single target:
  - Adapter is defined and holds one target. No queriying necessary
- Function is invoked through a code lens with multiple targets
  - Adapter is defined but holds multiple targets. we want to prompt the user for a target from the scope of the given targets.

Having said this, I would expect a structure like the following in the body of this function (or inside the new helper function `executeBazelCommand`):
- `if` statements to clearly separate these three usecases
- quickpicking for targets if required
- invocation of the actual command with the single identified target

What is important to me is that the control flow is as linear as possible, so that the code is both easy to read and understand, but also that we don't risk bugs by creating different behaviour in two different branches of the control flow.
The way it is currently implemented, I see that the logic can either take the `bazelMultipleTargetCommand` path or the `executeBazelCommand` path. Each of these has a quickpicking hidden inside and also a execution of the command, which is the contrary of what I stated before.

Hence I would propose to:
- Structure the helper function as proposed above
- Introduce a helper function called `selectSingleTarget` (or similar) which determines the usecase and performs the picking
- Move the implementation of the "new" quickpick logic from below (the one that will allow to quickpick from a given list of targets) into the  `bazel_quickpick.ts` file, to keep responsibilities clearly separated and to allow for reuse of similar code within this file
- Remove the `bazelMultipleTargetCommand` function in favor of the `executeBazelTask` function
- Avoid further `if` statements outside of the `selectSingleTarget` function to streamline controlflow
- Simplify the "entrypoints" for the commands, like `bazelBuildTarget` and `bazelTestTarget` which would then simply be two liners (select target & execute task) with the main responsibility of injecting the knowledge about which task to execute and how to name it (e.g. build vs test vs copy)
- Maybe even unify with all other functions in this file, like the ones for building packages (which would then simply add another usecase to the new select function)

cbandera (commented):
Thanks for adressing my last set of review comments. The functionality is still given and works fine locally.

Here is a next set of (opinionated) findings with the goal of improving code quality. 
(Note that I am not a maintainer, so feel free to take it as a suggestion or ignore as long as none of the maintainers performs a review 😉 )

src/codelens/bazel_build_code_lens_provider.ts
bkolb: we need to internally collect multiple code lenses for different categories. I think that is more readable.

bkolb:
OK. I pushed again. Let me know what you think. 
It would also be nice to get a review from @cameron-martin or any other maintainer :-)

cbandera:
Hi @bkolb do you mind updating your branch with the latest changes from upstream to resolve the merge conflicts? Thanks in advance!

cbandera:
I didn't test it again after the last merge with master. (And am on mobile only right now)
But primarily it's a review from one of the maintainers in order to get it merged...

cbandera:
Tested again locally and it still works fine. I'd be happy to see this in the next release :)

src/extension/bazel_wrapper_commands.ts
```diff
@@ -38,24 +93,17 @@ import { blaze_query } from "../protos";
  * which the command's arguments will be determined.
  */
 async function bazelBuildTarget(adapter: IBazelCommandAdapter | undefined) {
-  if (adapter === undefined) {
-    // If the command adapter was unspecified, it means this command is being
-    // invoked via the command palatte. Provide quickpick build targets for
-    // the user to choose from.
-    const quickPick = await showDynamicQuickPick({
```
cameron-martin, line 45: Not using `showDynamicQuickPick` any more here looks like a regression. Possibly an issue with merging with https://github.com/bazel-contrib/vscode-bazel/pull/465?

src/extension/bazel_wrapper_commands.ts
```diff
@@ -203,27 +257,17 @@ async function bazelRunTarget(adapter: IBazelCommandAdapter | undefined) {
...
-      queryFunctor: queryQuickPickTargets,
-      workspaceInfo: await BazelWorkspaceInfo.fromWorkspaceFolders(),
-    });
-    // If the result was undefined, the user cancelled the quick pick, so don't
-    // try again.
-    if (quickPick) {
-      await bazelTestTarget(quickPick);
-    }
-    return;
+  const selectedAdapter = await selectSingleTarget(
```
cameron-martin, line 258: Again here possibly the same regression.

src/extension/bazel_wrapper_commands.ts
cbandera: Good catch, I should have checked for that. The invocation has moved to L55 but is not using the new `showDynamicQuickPick`. Requires fixing.

src/extension/bazel_wrapper_commands.ts
cbandera: Yes, but requires fixing in the same location: L55

bkolb:
Ok. I can look at it. How can I test that it works? Is there a test for it?

bkolb:
Done. PTAL

cameron-martin:
> Done. PTAL

Please see the failing CI run.

cameron-martin (approved):

bkolb:
Thanks!

PR #481, opened 2025-10-17
fix(workspace-tree): honor bazel.commandLine.queryExpression for root targets

### LINKED ISSUE #480: `Bazel targets` does not honor `bazel.commandLine.queryExpression` for workspace root targets
Adding an expression to filter out workspace root targets such as `...:* except attr(name, \"^(\\.aspect_rules_js/|node_modules/).*\", //:*)` for example when using a workspace with `aspect_rules_js`, the Bazel targets list does not honor the `bazel.commandLine.queryExpression` for workspace root targets. Since these generated/shadow targets are all workspace root targets (without subpackages), the explorer view is drastically inflated.

MichaelSbarraKodiak:
- Respect the user's `bazel.commandLine.queryExpression` when querying workspace root targets by replacing `...:*` with `:all` if a workspace root BUILD file exists.
- Prevents explorer view bloat from generated/shadow root targets when using custom query expressions (e.g., with aspect_rules_js).
- Only queries root targets if the root `BUILD` or `BUILD.bazel` file is present.

Closes #480.

### DIFF
--- a/src/workspace-tree/bazel_workspace_folder_tree_item.ts
+++ b/src/workspace-tree/bazel_workspace_folder_tree_item.ts
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 import * as vscode from "vscode";
+import { existsSync } from "fs";
+import { join } from "path";
 import { BazelWorkspaceInfo, BazelQuery } from "../bazel";
 import { getDefaultBazelExecutablePath } from "../extension/configuration";
 import { blaze_query } from "../protos";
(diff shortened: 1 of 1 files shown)


src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -178,22 +180,30 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
...
-    const targets = queryResult.target.map((target: blaze_query.ITarget) => {
-      return new BazelTargetTreeItem(
-        this.resources,
-        this.workspaceInfo,
-        target,
-      );
-    });
+    let targets: BazelTargetTreeItem[] = [];
+    // Only get root targets if the root package is present
+    const rootBuildFile =
```
cbandera, line 185: How about reusing the `getBazelPackageFile()` method from `bazel_utils.ts`?

src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -178,22 +180,30 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
...
+      existsSync(join(workspacePath, "BUILD")) ||
+      existsSync(join(workspacePath, "BUILD.bazel"));
+
+    if (rootBuildFile) {
+      // Replace ...:* with :all in queryExpression
+      const queryExpressionRoot = queryExpression.replace(/\.{3}:\*/g, ':all');
```
cbandera, line 191: I am not sure whether I am missing something. Or maybe this changeset got corrupted by recent refactorings on main. But as far as I understand the code, there is no variable `queryExpression` here. Maybe you need to rebase onto latest `master`?

cbandera (commented):
Apart from the two findings in the code, I am not sure whether the suggested code change will actually do what you hope for.

The code in the modified function `getDirectoryItems()` will first collect all packages that it finds with the user specified `queryExpression`. And will initialize the target tree from this in `buildPackageTree`. After that, only the targets in the root package are queried and added to the tree.

This can be thought of as a lazy initialization of the tree. I.e. query all package names to quickly build a directory tree, but only fill in the details for those nodes of the tree that are expanded. (If you extend a node, the `getChildren` method will always query the targets for just this package). 

The code that you have modified does only query the toplevel targets during initial initialization. And with your change, it will now not only query the toplevel targets but potentially search for targets recursively within the specified scope. This is not correct here.

Having said that, I do have to admit however that the logic is already broken in case a user has specified a custom `queryExpression` that only queries a subsection of a repo, because in this case, the current logic would nevertheless query and display the toplevel targets anyway. (But maybe that's not too bad...)

PR #483, opened 2025-11-24
fix(buildifier): specify --lint=off in case buildifer config has different value

mattnworb:
buildifer allows you to configure defaults by placing a `.buildifer.json` file at the root of the project.

If this file happens to have

```json5
{
  "lint": "fix",
  ...
}
```

then the extension will run into an error at startup which presents a confusing "Failed to find a Bazel workspace" error dialog to the user. From the Extension Host logs in the Output panel:

```
2025-11-24 14:49:33.178 [error] Error: Command failed: buildifier --format=json --mode=check
buildifier: --lint=fix is only compatible with --mode=fix

	at genericNodeError (node:internal/errors:983:15)
	at wrappedFn (node:internal/errors:537:14)
	at ChildProcess.exithandler (node:child_process:418:12)
	at ChildProcess.emit (node:events:519:28)
	at maybeClose (node:internal/child_process:1101:16)
	at Socket.<anonymous> (node:internal/child_process:456:11)
	at Socket.emit (node:events:519:28)
	at Pipe.<anonymous> (node:net:346:12)
```

since the command in buildifer_availability is just to check that buildifier is present and able to return JSON output, the default from the config file can be overridden to prevent this error.

### DIFF
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -210,10 +210,12 @@ export function showDynamicQuickPick({
 
   let abortController: AbortController | undefined = new AbortController();
   let timeout: NodeJS.Timeout | undefined;
+  let statusBarMessage: vscode.Disposable | undefined;
+  let isValidating: boolean = false;
 
-  const updateQuickOptions = async (
-    currentUserInput: string,
-  ): Promise<void> => {
+  const executeQuery = async (
+    query: string,
+  ): Promise<BazelTargetQuickPick[] | undefined> => {
     // Cancel any previous query
     abortController?.abort();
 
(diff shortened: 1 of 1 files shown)


cameron-martin:
The buildifier help says the default is `--lint=off`. It looks like this will result in a change of behaviour.

cameron-martin (approved):
Thanks!

src/buildifier/buildifier_availability.ts
```diff
@@ -71,7 +71,9 @@ export async function checkBuildifierIsAvailable() {
   // output parses.
   const { stdout } = await executeBuildifier(
     "",
-    ["--format=json", "--mode=check"],
+    // specify the --lint value even though of is the default in case
```
cameron-martin:
```suggestion
    // specify the --lint value even though off is the default in case 
```

cameron-martin (commented):
Looks like there's linting errors

PR #484, opened 2025-11-27
fix(quickpick): improve loading state management in dynamic quickpick

cbandera:
Purpose: Improve loading / UX behavior of the quickpick dialog introduced in #465.

Behavioral fixes:
- Avoid removing quick-pick items while the query restarts after user input — the list now stays visible and interactive and is updated once the query finishes, eliminating the “come-and-go” effect.
- Accept user-typed labels even when no query results are available yet — if a user pastes/enters a full label and presses Enter before the initial query returns, a final validation/query will be executed to resolve that label.
- Show a cancellable progress message while the query is running to give the user feedback that a query is in progress (and to allow cancellation).

Note to the reviewer:
- Main functional changes are in the `showDynamicQuickPick` function.
- Two helper functions were extracted to allow for reuse and improve readability: `showProgressNotification`, `executeQuery`

### DIFF
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -210,10 +210,12 @@ export function showDynamicQuickPick({
 
   let abortController: AbortController | undefined = new AbortController();
   let timeout: NodeJS.Timeout | undefined;
+  let statusBarMessage: vscode.Disposable | undefined;
+  let isValidating: boolean = false;
 
-  const updateQuickOptions = async (
-    currentUserInput: string,
-  ): Promise<void> => {
+  const executeQuery = async (
+    query: string,
+  ): Promise<BazelTargetQuickPick[] | undefined> => {
     // Cancel any previous query
     abortController?.abort();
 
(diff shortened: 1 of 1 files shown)


src/bazel/bazel_quickpick.ts
```diff
@@ -210,21 +210,74 @@ export function showDynamicQuickPick({
...
+    if (resolveProgressNotification) {
+      resolveProgressNotification();
+      resolveProgressNotification = undefined;
+    }
 
+    if (isLoading) {
+      // Show progress in activity bar
+      vscode.window.withProgress(
+        {
+          location: vscode.ProgressLocation.Notification,
```
cameron-martin: My first impression when using this is that this type of progress notification is too intrusive for this context. `vscode.ProgressLocation.Window` seems much nicer to me. What do you think?

I'm also not too convinced that a second loading indicator is needed, since we already have one in the quick pick.

cbandera: You are absolutely right. I just wasn't aware of the different possibilities to show the status.
I have reworked the progress messages and made sure that the quickpick stayed busy until the results are available.

cameron-martin (approved):
Thanks

PR #487, opened 2025-12-06
fix(buildifier): Support linting & formatting MODULE.bazel files

### LINKED ISSUE #471: Correctly format Module.bazel files with buildifier
Now plugin runs buildifier with `--type=default` for `Module.bazel` files, but correct type is `module`. 

From buildifier doc
```
-type string
        Input file type: build (for BUILD files), bzl (for .bzl files), workspace (for WORKSPACE files), module (for MODULE.bazel files), default (for generic Starlark files) or auto (default, based on the filename) (default "auto")
```

https://github.com/bazelbuild/buildtools/blob/4006b543a694f6cf77c2d8cc188c5f53f3bac1d9/build/lex.go#L125

cameron-martin:
This bug was caused by the logic for computing the file type to pass to buildifier not supporting module.bazel files. However, support for inferring the file type based on the passed-in file path was added to buildifier back in 2019, so I've updated it to use this.

I also added a test to ensure the diagnostics are being added correctly. This requires buildifier to be installed on CI so I've added an action to install this too.

Fixes #471

### DIFF
--- a/test/workspace_tree.test.ts
+++ b/test/workspace_tree.test.ts
@@ -77,6 +77,7 @@ describe("Bazel Workspace Tree", function (this: Mocha.Suite) {
   it("should match workspace structure", async () => {
     await verifyTreeStructure(
       {
+        "//buildifier": {},
         "//pkg1": {
           ":foo  (filegroup)": {},
           ":main  (py_binary)": {},
@@ -103,6 +104,7 @@ describe("Bazel Workspace Tree", function (this: Mocha.Suite) {
     // THEN
     await verifyTreeStructure(
       {
+        "//buildifier": {},
         "//pkg1": {
           ":foo  (filegroup)": {},
           ":main  (py_binary)": {},


src/buildifier/buildifier.ts
```diff
@@ -37,18 +37,18 @@ export type BuildifierFileType = "build" | "bzl" | "workspace" | "default";
...
  * @returns The formatted file content.
  */
 export async function buildifierFormat(
   fileContent: string,
-  type: BuildifierFileType,
+  filePath: string,
   applyLintFixes: boolean,
 ): Promise<string> {
-  const args = [`--mode=fix`, `--type=${type}`];
+  const args = [`--mode=fix`, `--path=${filePath}`];
```
cbandera, line 51: The docs state that this argument has only one leading dash. https://github.com/bazelbuild/buildtools/blob/main/buildifier/buildifier.go#L66-L70
But it seems to accept both.

src/buildifier/buildifier.ts
```diff
@@ -78,26 +78,26 @@ export async function buildifierLint(
...
- * .bzl file.
+ * @param filePath The path to the file being formatted, relative to the
+ * workspace root.
  * @param lintMode Indicates whether to warn about lint findings or fix them.
  * @returns An array of objects representing the lint issues that occurred.
  */
 export async function buildifierLint(
   fileContent: string,
-  type: BuildifierFileType,
+  filePath: string,
```
cbandera, line 88: While reading through the docs, I noticed that `buildifier` also supports the `--mode=diff` mode, which would only print the differences, that the formatter would apply.

Considering the API of VSCode, where our formatter returns only the edits, I feel like this could be (didn't think it through completely yet) a cleaner approach.
Benefits that I see are:
- no need to read the complete file contents and pass them on stdin again
- no need to pass `--path` since we would pass the real filename anyway
- maybe performance gains for big files
- in case no changes are necessary, we would not indicate to VScode a necessary change. The way I understand the current implementation, the complete file contents are replaced every time at the moment.

Let me know whether you think this change is worth exploring.

cbandera (commented):
Changes look and work fine. Just two comments for discussion

src/buildifier/buildifier.ts
cameron-martin: I saw that, but we're passing double slash everywhere else. Do you think we should change them all to use a single slash?

src/buildifier/buildifier.ts
cameron-martin: I presume you're referring to the buildifierFormat function, even though this comment is on the buildifierLint function.

Using `mode=diff` does sound like an interesting approach. I think you'd still have to write the file via stdin though, because the file contents in the editor isn't necessarily the same as the file contents on disk.

src/buildifier/buildifier.ts
cbandera: No, it's "proven in use". It was more that I was wondering why it works, because I saw no handling of double-dashes in the source code. But it's probably hidden in some helper library.

src/buildifier/buildifier.ts
cbandera: Ah, is see. Good point. I might have a look at the diff option separately though

PR #489, opened 2025-12-09
fix(config): Make default values for executables user-visible

### LINKED ISSUE #88: getDefaultBazelExecutablePath() returns empty string

cbandera:
This change makes the default settings for the executable paths visible from the VSCode GUI and simplifies the code by moving the defaults to `package.json`.

Closes #88

### DIFF
--- a/src/buildifier/buildifier.ts
+++ b/src/buildifier/buildifier.ts
@@ -127,7 +127,7 @@ export function getDefaultBuildifierExecutablePath(): string {
   return (
     vscode.workspace
       .getConfiguration("bazel")
-      .get<string>("buildifierExecutkable")
+      .get<string>("buildifierExecutable")
       .trim() || "buildifier"
   );
 }


cameron-martin:
After this change if you set the settings to the empty string it fails to spawn bazel or buildifier. Since the empty string is never a valid executable, maybe we should keep the behaviour where we set a default value when the setting is the empty string?

src/buildifier/buildifier.ts
```diff
@@ -118,33 +118,32 @@ export async function buildifierLint(
...
-  const bazelConfig = vscode.workspace.getConfiguration("bazel");
-  const buildifierExecutable = bazelConfig.get<string>("buildifierExecutable");
-  if (buildifierExecutable.length === 0) {
-    return "buildifier";
-  }
-  return buildifierExecutable;
+  return (
+    vscode.workspace
+      .getConfiguration("bazel")
+      .get<string>("buildifierExecutkable")
```
cameron-martin: Looks like you've got a typo in this key. Possibly what's causing the test failures.

cameron-martin:
```suggestion
      .get<string>("buildifierExecutable")
```

PR #491, opened 2025-12-11
test: Make buildifier test deterministic

cbandera:
This PR makes the buildifier test deterministic by:
- awaiting for diagnostics to be produced before asserting
- safeguarding against endless runs with a timeout
- do not check exact match, but assert that the warning of interest is among the diagnostic warnings

Adresses: https://github.com/bazel-contrib/vscode-bazel/pull/487#issuecomment-3634589309

### DIFF
--- a/test/buildifier.test.ts
+++ b/test/buildifier.test.ts
@@ -29,15 +29,24 @@ const workspacePath = path.join(
 
 describe("buildifier", () => {
   it("diagnostics are added from buildifier", async () => {
+    // Create DiagnosticsManager and open file
     const buildFile = path.join(workspacePath, "buildifier", "BUILD");
     await openSourceFile(buildFile);
-
     disposables.push(new BuildifierDiagnosticsManager());
 
-    const diagnostics = vscode.languages.getDiagnostics(
-      vscode.Uri.file(buildFile),
-    );
+    // Promise that resolves when diagnostics are added to the file.
+    const diagnosticsPromise = new Promise<vscode.Diagnostic[]>((resolve) => {
+      const disposable = vscode.languages.onDidChangeDiagnostics((e) => {
+        const uri = vscode.Uri.file(buildFile);
+        const diags = vscode.languages.getDiagnostics(uri);
+        disposable.dispose();
+        resolve(diags.slice(0, 1)); // Return first diagnostic only
+      });
+    });
+    // Wait for diagnostics for this file
+    const diagnostics = await diagnosticsPromise;
 
+    // Assert the diagnostic matches expectation
     assert.deepEqual(diagnostics, [
       {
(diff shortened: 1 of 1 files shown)


test/buildifier.test.ts
```diff
@@ -29,32 +29,55 @@ const workspacePath = path.join(
...
 describe("buildifier", () => {
   it("diagnostics are added from buildifier", async () => {
+
+    // Promise that resolves when diagnostics are added to the file.
+    const diagnosticsPromise = new Promise<vscode.Diagnostic[]>(
+      (resolve, reject) => {
+        const timeout = setTimeout(() => {
+          disposable.dispose();
+          reject(new Error("Timed out waiting for diagnostics"));
+        }, 5000);
```
cbandera: I guess this timeout could also be removed since the tests have their own timeout. What do you think?

cbandera: I simplified the code..

test/buildifier.test.ts
```diff
@@ -29,32 +29,54 @@ const workspacePath = path.join(
...
 
-    disposables.push(new BuildifierDiagnosticsManager());
+    // Wait for diagnostics for this file
+    const diagnostics = await diagnosticsPromise;
 
-    const diagnostics = vscode.languages.getDiagnostics(
-      vscode.Uri.file(buildFile),
-    );
+    // Assert expected diagnostic is among returned diagnostics
+    const hasExpected = diagnostics.some((d) => {
```
cameron-martin: How come you've refactored the assertion like this? This will give much worse error messages when the assertion fails.

test/buildifier.test.ts
cameron-martin: Yeah I think relying on the test timeout is fine here, and would simplify the test a bit.

test/buildifier.test.ts
cbandera: The reason was that, while debugging, I saw that there were actually multiple diagnostic messages in the `diagnostics` object. I think it was simply the same diagnostic message twice. 
So by refactoring as I did, I now check that at least one of the contained diagnostics matches the expectation which should make it more robust.
I also stepped away from the deep equality check as I at some point thought that the problem was that the diagnostics didn't match 100%.... but I will try to revert this. should still work with the old way of asserting

cbandera: Ok, I revisited the code and kept my refactoring to a minimum so that the original idea of the test is kept, but at the same time the robustness is increased.

cbandera:
Maybe one of the other maintainers has time to approve this PR? It is currently causing the CI to fail on all other PRs...
Thanks @jfirebaugh , @vogelsgesang , @coeuvre

Issue #492, opened 2025-12-11
Version 0.13.0 opens file explorer when focused tab changes


ingu119:
Whenever you switch tabs in VSC, the file explorer opens and highlights the current file, which is very annoying. This issue only arises with Bazel extension version 0.13.0. Could it be that this is a bug in this extension/version?

cbandera:
I have created a quick fix for this which solves the issue for me. Would one of you be willing to test the new version and confirm that your respective issue is solved as well?
You can download the .vsix file from the CI run: https://github.com/bazel-contrib/vscode-bazel/actions/runs/20131946546?pr=493
To install it in your VSCode you can select "Extensions: Install from VSIX..." from the command palette and select the file. You probably need to "Reload Window" afterwards to take effect. Thanks in advance!

PR #493, opened 2025-12-11
fix(worskpace-tree): Do not reveal treeview when not visible.

cbandera:
Closes: #492


test/workspace_tree.test.ts
cbandera, line 1: I had a hard time writing a positive test for this, because it does not seem to be possible to programatically "expand" the bazel target pane from within a test case. As such, the target tree is always `not visible`. 
But as this is a quick fix, I would prefer to merge as-is and rather ship a version `0.13.1` as quick as possible

PR #496, opened 2025-12-16
feat: Allow to deactivate CodeLens dynamically

cbandera:
This PR is now a followup of #589 and turns the CodeLens feature into the first ExtensionFeature.

While the config setting for enabling CodeLens does exist already, this mechanism allows us to dynamically enable and disable this features during runtime, without the need to `Reload Window` or re-activation of the whole extension. It also takes care of proper logging, error messaging and precondition-checking.

You can playaround with it by opening a build file and the settings in split view and toggle the `bazel.enableCodeLens` setting.
Also, if you configure a non-existent path for the `bazel.executable`, you can see the error-path.

As for the design: I have spread the implementation across 3 files, as I believe that this makes each of them much easier to understand (and test) then in the previous implementation:
- `CodeLensFeature` is the implementation of the `BaseExtensionFeature` class and takes care of checking preconditions before enabling the feature as well as registering vscode-related resources.
- `CodeLensProvider` is the implementation of the `vscode` API for code lenses and interacts with `Bazel` to gather info on which code lenses to build.
- `CodeLensBuilder` finally contains all the businesslogic for deciding what kind of interactive buttons the codelens should have and what actions to trigger.


src/codelens/code_lens_builder.ts
```diff
@@ -251,14 +153,6 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
 
   /**
    * Creates a CodeLens for a specific action type if targets are available.
-   * Title shows action name with count for multiple targets.
```
cameron-martin, line 254: Why have these comments been removed?

cbandera:
@cameron-martin I have adapted this PR to the logger changes on main. Have you had a chance to look at the proposed "Extension Feature" yet? I'd love to hear your critical thoughts on it

cameron-martin:
> @cameron-martin I will postpone this PR to 0.15, so that we can bundle a 0.14 release soon. But it would be great if you could do this review at some point so that I can continue rebuilding the other features to follow this pattern as well.

I've been meaning to have a look at this but have failed to find the time recently. Bear with me but yes doing the 0.14 release first sounds reasonable.

src/codelens/code_lens_feature.ts
```diff
@@ -0,0 +1,61 @@
...
+ * - Provider registration with VSCode
+ */
+export class CodeLensFeature extends BaseExtensionFeature {
+  constructor(context: vscode.ExtensionContext) {
+    super("CodeLens", context);
+  }
+
+  enable(context: vscode.ExtensionContext): boolean {
+    // Precondition: must be in a Bazel workspace
+    if (!checkBazelWorkspaceAvailable()) {
```
cbandera: I am not so sure anymore whether this is a good idea: if I were to open a non Bazel Workspace, I would get verbose error messages, warning me that this feature cannot be activated. 
Instead, this case should be handled gracefully and silently

cbandera: Addressed.

cbandera:
@cameron-martin  this PR is ready for review if you find the time :)

src/codelens/code_lens_builder.ts
```diff
@@ -205,10 +116,6 @@ export class BazelBuildCodeLensProvider implements vscode.CodeLensProvider {
 
   /**
    * Groups targets by the actions they support based on Bazel rule types.
-   * Uses rule naming conventions to determine which actions are available.
```
cameron-martin, line 208: How come these comments are removed?

src/codelens/code_lens_builder.ts
```diff
@@ -13,22 +13,12 @@
...
-
 import { BazelWorkspaceInfo, QueryLocation } from "../bazel";
-import { getTargetsForBuildFile } from "../bazel";
-import { getBazelExecutablePath } from "../extension/configuration";
 import { blaze_query } from "../protos";
 import { CodeLensCommandAdapter } from "./code_lens_command_adapter";
 
 /**
  * Groups of Bazel targets organized by the actions they support.
- * Used by the CodeLens provider to determine which actions to display for each target.
```
cameron-martin, line 25: These comments look useful. How come they're removed?

cameron-martin (changes requested):
Looks good, but it looks like we may have lost useful comments in the process!

cbandera:
@cameron-martin I have brought back the comments. I think I removed them because I thought that they were to verbose. But I don't have a problem keeping them if you consider them useful. This also makes the diff much cleaner.

cameron-martin (approved):

PR #506, opened 2026-01-06
feat(navigation): Improve "Go to BUILD file" command

### LINKED ISSUE #502: Bug/Feature Request: Place cursor appropriately when calling `Go to BUILD file`
Currently when you use `Go to BUILD file`, the cursor will be located at the last touched location in the BUILD file that opens.  Instead, the cursor should be placed at the reference to the current source file.

cbandera:
The command will now attempt to jump to the correct line inside the build file, based on a simple search for a reference of the source file.

fixes #502

### DIFF
--- a/src/extension/bazel_wrapper_commands.ts
+++ b/src/extension/bazel_wrapper_commands.ts
@@ -22,6 +22,7 @@ import {
   getBazelPackageFile,
   getBazelWorkspaceFolder,
   getBazelPackageFolder,
+  getBuildFileLineWithSourceFilePath,
 } from "../bazel/bazel_utils";
 import {
   queryQuickPickTargets,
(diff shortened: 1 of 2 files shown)


src/extension/bazel_wrapper_commands.ts
```diff
@@ -373,7 +375,24 @@ async function bazelGoToBuildFile() {
...
+      // Move cursor to the line with the reference it found
+      const position = new vscode.Position(i, 0);
+      editor.selection = new vscode.Selection(position, position);
+      editor.revealRange(
+        new vscode.Range(position, position),
+        vscode.TextEditorRevealType.InCenter,
+      );
+      break;
+    }
+  }
```
cbandera, line 382: Optional: Should we issue a message to the user if we weren't successfull in finding the correct line?
Or is that just noise?

src/extension/bazel_wrapper_commands.ts
```diff
@@ -368,8 +370,24 @@ async function bazelGoToBuildFile() {
   }
 
   // Open the BUILD file
+  const editor = await vscode.window.showTextDocument(
+    vscode.Uri.file(buildFilePath),
+  );
 
-  await vscode.window.showTextDocument(vscode.Uri.file(buildFilePath));
+  // Find the line number where the current file is referenced
```
cbandera: Can reuse `getBuildFileLineWithSourceFilePath` once #543 is merged.

src/extension/bazel_wrapper_commands.ts
cameron-martin: I think that's just noise. It's not like something has gone wrong because finding the location is only best-effort.

src/extension/bazel_wrapper_commands.ts
cbandera: Done

PR #542, opened 2026-01-10
perf(workspace-tree): implement debouncing and lazy loading for performance 

### LINKED ISSUE #504: Bug: hitting process limit due to `bazel query ...:*` spawning from extension
I've found that on `1.13.1`, I'm able to consistently repro an issue where the extension spawns 1000s of `bazel query ...:*` due to some code tools that create/modify/delete 1000s of bazel build files (codegenned bazel, etc.). This doesn't happen on `1.12.0` and I'd only see the bazel processes tick up by at most 8 (including some other ongoing codegen, etc.).

Nothing seemed obvious among the code changes that would cause this - the mostly likely thing I could see would be the dynamic quickpick which modifyed the bazel query execution but I couldn't spot the bug.

As some issue SEO, this would result in the following error on when attempting to do anything in the terminal due to being unable to create another process: `fork failed: resource temporarily unavailable`

cbandera:
- Add 500ms debouncing for BUILD file changes to prevent refresh storms
- Implement lazy loading to skip refreshes when tree view is hidden
- Queue refreshes when tree is hidden and execute when visible again
- Ensure that only one Tree refresh is ever running at once
- Remove duplicate initialization for TreeItems (init not needed in constructor)

This significantly improves performance during bulk operations like git checkout by reducing hundreds of immediate refreshes to a single debounced refresh.

Fixes #504

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Lint
         run: npm run lint-check
 
-      - uses: bazel-contrib/setup-bazel@0.16.0
+      - uses: bazel-contrib/setup-bazel@0.18.0
         with:
           bazelisk-cache: true
           disk-cache: false
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
                 "@vscode/vsce": "^3.0.0",
                 "eslint": "^9.0.0",
                 "eslint-config-prettier": "^10.0.0",
-                "eslint-plugin-jsdoc": "^61.0.0",
+                "eslint-plugin-jsdoc": "^62.0.0",
                 "js-yaml": "^4.1.1",
                 "mocha": "^11.0.0",
                 "prettier": "^3.2.5",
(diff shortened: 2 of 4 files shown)


src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -132,22 +136,51 @@ export class BazelWorkspaceTreeProvider
...
+    // Clear any existing timeout
+    if (this.refreshTimeout) {
+      clearTimeout(this.refreshTimeout);
+    }
+
+    // Set a timeout to perform refresh after a short delay
+    this.refreshTimeout = setTimeout(() => {
+      void (async () => {
+        this.refreshTimeout = undefined;
+        await this.refresh();
```
Silic0nS0ldier: `this.refresh()` is not an `async` function nor does it return a promise. `await` is unnecessary.

src/workspace-tree/bazel_workspace_tree_provider.ts
```diff
@@ -191,6 +230,24 @@ export class BazelWorkspaceTreeProvider
    */
   public setTreeView(treeView: vscode.TreeView<IBazelTreeItem>): void {
     this.treeView = treeView;
+
+    // Set up visibility change handler for lazy loading
+    this.disposables.push(
+      this.treeView.onDidChangeVisibility(async () => {
+        await this.onTreeViewVisibilityChanged();
```
Silic0nS0ldier: `this.onTreeViewVisibilityChanged()` is not an `async` function nor does it return a promise. `await` is unnecessary.

src/workspace-tree/bazel_workspace_tree_provider.ts
cbandera: Right, good catch. Will fix

src/workspace-tree/bazel_workspace_tree_provider.ts
cbandera: Right. Fixed

PR #550, opened 2026-01-24
ci: Combine build and release workflows

cameron-martin:
Due to limitations of the GH_TOKEN, where actions caused by it do not trigger new workflows, the build artifacts were not being uploaded to releases, since this relied on CI running when a release was made.

Now the same workflow that creates the release uploads the release artifact. To make these build artifacts available to the release workflow, everything has been combined into one `build-release` workflow.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -314,19 +314,41 @@
             }
         },
         "node_modules/@es-joy/jsdoccomment": {
-            "version": "0.79.0",
-            "resolved": "https://registry.npmjs.org/@es-joy/jsdoccomment/-/jsdoccomment-0.79.0.tgz",
-            "integrity": "sha512-q/Nc241VsVRC5b1dgbsOI0fnWfrb1S9sdceFewpDHto4+4r2o6SSCpcY+Z+EdLdMPN6Nsj/PjlPcKag6WbU6XQ==",
+            "version": "0.82.0",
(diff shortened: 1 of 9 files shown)


.github/workflows/build-release.yml
```diff
@@ -2,13 +2,13 @@ name: Build VS Code extension
 
 on:
   push:
+    branches: [main]
```
cameron-martin: This restriction avoids the duplicate workflow runs we were seeing on PRs

cbandera: Shouldn't this be `master` though?

.github/workflows/build-release.yml
```diff
@@ -74,11 +74,16 @@ jobs:
           path: vscode-bazel-*.vsix
           if-no-files-found: error
 
+      - uses: googleapis/release-please-action@v4
+        id: release
```
cbandera, line 78: Shouldn't we add a conditional to restrict that this only runs when the trigger was a push to master (instead of on PR runs) ~~and only if this is executed inside the main `bazel-contrib` repo~~ (edit: happening at job level already)

.github/workflows/build-release.yml
cameron-martin: Yes you're right

.github/workflows/build-release.yml
cameron-martin: Yes you're right, I've added this conditional

PR #552, opened 2026-01-24
refactor: Make codebase more compatible with strictNullChecks

cameron-martin:
With the goal to enable the typescript `strictNullChecks` option in the near future, this fixes some places that would error when `strictNullChecks` is enabled. Specifically this reduces the errors from 167 to 96. Note this does not actually enable this typescript option because many errors still remain.


src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -193,18 +196,14 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
...
     const workspacePath = this.workspaceInfo.workspaceFolder.uri.fsPath;
     const packagePaths = await new BazelQuery(
       getDefaultBazelExecutablePath(),
       workspacePath,
-    ).queryPackages(
-      vscode.workspace
-        .getConfiguration("bazel.commandLine")
-        .get("queryExpression"),
-    );
+    ).queryPackages(getQueryExpression());
```
cbandera, line 206: It's probably not the primary goal of your PR to unify all occurances... but there is one more occurance that could be adapted to use the new getter inside `bazel_quickpick`

src/extension/configuration.ts
```diff
@@ -26,6 +26,30 @@ export function getDefaultBazelExecutablePath(): string {
...
+      .getConfiguration("bazel.commandLine")
+      .get<string[]>("commandArgs") ?? []
+  );
+}
+
+export function getQueryExpression(): string {
+  return (
+    vscode.workspace
+      .getConfiguration("bazel.commandLine")
+      .get<string>("queryExpression") ?? "...:*"
```
cbandera: The defaults that are given in `package.json` are reapeted here. 
Is there a way to avoid this repetition and rely on the fact that the value should always be set? If not, can we ensure that the defaults stay in sync? Maybe via a test?

src/workspace-tree/bazel_workspace_folder_tree_item.ts
cameron-martin: I don't see that, nor can I find it by searching. What line are you referring to?

src/extension/configuration.ts
cameron-martin: I've made this read the default from the package.json now, and extracted this logic out into a separate function.

src/workspace-tree/bazel_workspace_folder_tree_item.ts
cbandera: I saw this occurrence:
https://github.com/bazel-contrib/vscode-bazel/blob/master/src/bazel/bazel_quickpick.ts#L266

cameron-martin: That occurrence is already updated to use `getQueryExpression`

cbandera: I am sorry. I must have mixed up branches while I was reviewing. :S

src/extension/configuration.ts
```diff
@@ -14,6 +14,31 @@
...
 import * as vscode from "vscode";
 
+/**
+ * Gets a configuration value, returning the default set in the package.json if
+ * the value is not set or falsey.
+ * @param section The section that contains the configuration, as is passed to vscode.workspace.getConfiguration.
+ * @param name The name of the configuration item within the section.
+ * @returns The configuration value, or its default.
+ */
+function getConfigurationWithDefault<T>(section: string, name: string): T {
```
cbandera: If you find the time: Can you also add a unittest for this new function?

cbandera (approved):

src/extension/configuration.ts
cameron-martin: On second thoughts I don't think this is necessary. I don't think it's possible for a configuration option to return undefined if there is a default. I'll update this to just assert that the value isn't undefined.

cbandera: That's fine as well. But still a unittest would proof that we don't need to fear of this not working in the default case.

Issue #538, opened 2026-01-08
Bundle extension during packaging step


cbandera:
Recently we get this warning during the Packaging step in CI

> Warning: This extension consists of 3043 files, out of which 2244 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension. You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore.

This should be addressed.

PR #553, opened 2026-01-25
build: Cherry-pick improvements from #421

cbandera:
This PR takes over some of the non-esbuild related improvements from #421. 

Related to #538


scripts/prepare.sh
cameron-martin, line 1: What's the reasoning behind extracting this script from build.sh? On the surface it seems like a net negative, because:

* You can no longer run `./scripts/build.sh` to run the whole build.
* The dependency between prepare.sh and build.sh (that the former has to be run first) is expressed in multiple places now in the package.json.

Issue #198, opened 2020-05-23
Refreshing 'Bazel Build Targets' fails when there is a problem in the build graph


bholmes-sc:
Refreshing a the test of build targets tries to pull every target in the workspace using `bazelisk query '//...:*' --output=package`.  If any target in the graph fails then the targets are unable to refresh.

This can be fixed by adding `--keep_going` to the command.

thinlizzy:
this is happening to me as well. every change that I make in the BUILD files makes the targets no longer "refreshable". 

even after I fix the BUILD files, I have to manually delete the local bazel cache and then they I can refresh the targets again

derekwisong:
I've also found the same. If I run `bazel clean --expunge` and then refresh the build targets it works. As soon as I run a build, the refresh targets stops working again... Until I do another `bazel clean --expunge` (just doing `bazel clean` isn't enough).

cameron-martin:
@derekwisong See #216

PR #556, opened 2026-01-26
Add bazel query options setting 

markkohdev:
## Description
For very large bazel repos, [git sparse checkout](https://git-scm.com/docs/git-sparse-checkout) is commonly used to only pull a subset of directories in a repo.  When this happens, the local version of the repo may not have all of the dependencies of all packages or tools required and this is expected.

Because the current `bazel query` call on extension startup passes `...:*` as the default query, the extension will automatically try to query all packages and will fail if any are missing, and will infinitely re-query and fail.  This can be resolved by adding `--keep_going` to the query command.

This PR adds a `bazel query packages options` setting to allow users to append custom parameters to the bazel query to support things such as sparse checkout.

## Related Issue
#198 

## Testing
- [x] Tests added/updated

## Checklist

### DIFF
--- a/src/bazel/bazel_query.ts
+++ b/src/bazel/bazel_query.ts
@@ -20,7 +20,7 @@ import * as vscode from "vscode";
 import { blaze_query } from "../protos";
 import { BazelCommand } from "./bazel_command";
 import { getBazelWorkspaceFolder } from "./bazel_utils";
-import { logDebug, logError } from "../extension/logger";
+import { logDebug, logError, logWarn } from "../extension/logger";
 
 const protoOutputOptions = [
   "--proto:output_rule_attrs=''",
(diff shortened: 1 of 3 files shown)


src/bazel/bazel_query.ts
```diff
@@ -58,8 +58,12 @@ export class BazelQuery extends BazelCommand {
       abortSignal?: AbortSignal;
     } = {},
   ): Promise<blaze_query.QueryResult> {
+    const bazelConfig = vscode.workspace.getConfiguration("bazel");
```
cbandera, line 61: Once #552 is merged, please adapt and move to `configuration.ts`

cbandera (commented):
I wonder whether we shouldn't set the `--keep_going` flag by default, in order to be robust against temporary changes. 
I.e. I keep seeing errors when I save my build file with partial changes and my build tree hence becomes temporarily corrupt....

cbandera:
If the scope of the problem that you are facing in your setup is really "just" the querying for packages (which is only used for the workspace-tree) then I would prefer to simply pass --keep_going for this specific query by default.
With the new logger we do have the possibility to point the user to more information in the log output.

Adding too many configuration options for the users that most of the devs wont ever use, just clutters the settings and makes it harder to use.

I would also vote against coloring the workspace tree - that's too much added complexity.

@cameron-martin do you have any thoughts on this PRs proposal?

cameron-martin:
I agree that always passing `--keep_going` sounds reasonable instead of adding another option.

PR #557, opened 2026-01-26
feat(logging): Add proper VSCode Logging Output

### LINKED ISSUE #227: bazel query getting stuck

markkohdev:
## Description

This PR introduces a centralized logging module to replace direct `console.log`/`console.error` calls and scattered `vscode.window.show*Message` usage throughout the extension. The new logger provides:

- **Structured logging** via VS Code's `LogOutputChannel` with proper log levels (error, warn, info, debug)
- **Optional user-facing messages** with a "Details" button that opens the output channel for full context
- **Consistent error formatting** with support for additional arguments using `util.format`
- **Better debugging experience** by centralizing all logs in a single output channel

### Key Changes

- Added `src/extension/logger.ts` with logging API (`logError`, `logWarn`, `logInfo`, `logDebug`)
- Refactored 15+ files to use the new logger instead of direct console/window message calls
- Improved error handling in `bazel_query.ts` (better error formatting, fixed duplicate query error)
- Enhanced `bazel_utils.ts` error handling for invalid regex patterns and missing files
- Added comprehensive test suite (`test/logger.test.ts`) with 573 lines covering all logging scenarios

### Usage Examples

**Setup (in extension activation):**
```typescript
import { registerLogger, logInfo, logError } from "./extension/logger";

export async function activate(context: vscode.ExtensionContext) {
  registerLogger("Bazel VSCode", context);
  logInfo("Extension activated successfully.");
  // ...
}
```

**Basic logging (output channel only):**
```typescript
import { logDebug, logInfo, logWarn, logError } from "./extension/logger";

logDebug("Running Bazel query", false, `Command: ${command}`);
logInfo("Build completed successfully");
logWarn("Buildifier not found", false);
logError("Query failed", false, error);
### Usage Examples

**Setup (in extension activation):**
```typescript
import { registerLogger, logInfo, logError } from "./extension/logger";

export async function activate(context: vscode.ExtensionContext) {
  registerLogger("Bazel VSCode", context);
  logInfo("Extension activated successfully.");
  // ...
}
```

**Basic logging (output channel only):**
```typescript
import { logDebug, logInfo, logWarn, logError } from "./extension/logger";

logDebug("Running Bazel query", false, `Command: ${command}`);
logInfo("Build completed successfully");
logWarn("Buildifier not found", false);
logError("Query failed", false, "Error details:", error);
logError("Package discovery failed", false, "Unable to find file: %s", filepath);
```

**User-facing messages with details:**
```typescript
// Shows error notification with "Details" button
logError("Failed to find Bazel workspace", true);

// Shows info notification + logs full details to output channel
logInfo("Bazel build completed successfully in 5 seconds.", true, buildOutputStr);

// Shows warning + logs additional context
logWarn(
  "Buildifier not available; linting disabled.",
  true,
  "Download from:",
  BUILDTOOLS_RELEASES_URL
);
```

**Error handling with context:**
```typescript
try {
  await bazelQuery.execute();
} catch (error) {
  logError(
    "Bazel query failed",
    true,  // Show user notification
    "Query command output:",
    errorOutput,
    "Exit code:",
    code
  );
}
```

### Additional Improvements

- Added PR template (`.github/pull_request_template.md`) for consistent PR descriptions
- Updated `CONTRIBUTING.md` with clearer testing instructions
- Added `npm run clean` script for easier test environment cleanup
- Removed `out/` dir from vscode file exclusion to facilitate easier test debugging

## Related Issue

Closes #227 

## Testing

- [x] Tests added/updated
  - Added comprehensive unit tests for logger module (573 lines)
  - Tests cover all log levels, user messaging, argument formatting, and edge cases
  - All existing tests continue to pass
  - Verified logging output appears correctly in VS Code output channel
  - Confirmed user messages display with "Details" button functionality
  - Tested error scenarios to ensure proper logging and user notifications

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,16 +1,21 @@
 ## Description
+
 <!-- Brief description of changes, screenshots if applicable -->
 
 ## Related Issue
+
 <!-- Link to issue or mention if claiming an existing issue -->
 <!-- e.g: Closes #123 -->
 
 ## Testing
+
 <!-- Describe how you tested your changes -->
+
 - [ ] Tests added/updated
 - [ ] Manual testing performed
 
 ## Checklist
+
 - [ ] Code follows the project's style guidelines (prettier/eslint)
 - [ ] Commit messages follow [Conventional Commit](https://www.conventionalcommits.org/) conventions
 - [ ] Tests pass locally (`npm run test`)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ extensions to have these tools applied automatically as you develop:
 
 We expect contributions to include tests that demonstrate and validate the intended behavior.
 
-- Running tests: Use `npm run test` to run the tests.  These tests will download a local version of vscode into the `.vscode-test` directory and will run integration tests against the extension in that workspace.
+- Running tests: Use `npm run test` to run the tests. These tests will download a local version of vscode into the `.vscode-test` directory and will run integration tests against the extension in that workspace.
 
 - New features: add a set of VS Code integration tests that demo the intended behaviour by making use of the mock workspace under `test/bazel_workspace` (vscode-tests will be executed within this workspace). Integration tests should exercise the extension as a user would (for example: open editors, execute commands with `vscode.commands.executeCommand`, and assert on the editor state or extension outputs). See `test/go_to_label.test.ts` for a minimal integration-style example.
 


.vscode/settings.json
```diff
@@ -1,7 +1,6 @@
 {
     "files.exclude": {
-        "node_modules": true,
-        "out": true
```
markkohdev, line 4: Note: Excluding the `out` directory from the `files` panel made testing very difficult since there wasn't visibility into stale files (which I needed to clean up with the new `npm run clean` command).  If there were particular reasons that this was excluded, then we should use a more targeted approach than just `files.exclude`

src/bazel/bazel_utils.ts
```diff
@@ -111,8 +111,13 @@ function findAncestorFile(
   let iteration = 0;
   const maxIterations = 100; // Fail-safe to prevent infinite loops
 
-  if (fs.statSync(startPath).isFile()) {
-    dirname = path.dirname(dirname);
+  try {
```
markkohdev, line 118: Macs don't have some of the files that are searched for during quickpick tests ("/etc/hostname"), causing crashing when trying to run tests locally.

cbandera: I wonder whether we shouldn't rather adapt the testcase then

src/extension/logger.ts
```diff
@@ -0,0 +1,275 @@
...
+// https://github.com/microsoft/vscode-black-formatter/blob/main/src/common/logging.ts
+import * as util from "util";
+import * as vscode from "vscode";
+import { Disposable, ExtensionContext, LogOutputChannel } from "vscode";
+
+type Arguments = unknown[];
+
+const DETAILS_ACTION = "Details";
+let channel: LogOutputChannel | undefined;
+let outputChannelName: string | undefined;
```
cbandera: I would prefer to not make the channel name configurable but hardcode it here.
Also I would call it `Bazel` instead of `Bazel Vscode` (the vscode in the repo name is really just convention for naming extensions... but user facing, `Bazel` is what makes most sense

src/extension/logger.ts
```diff
@@ -0,0 +1,275 @@
...
+      log: true,
+    },
+  );
+  context.subscriptions.push(outputChannel);
+
+  outputChannelName = name;
+  channel = outputChannel;
+  channel.info("Logger initialized.");
+
+  const loggerDisposable: Disposable = {
```
cbandera, line 56: `createOutputChannel` already returns a `dispose`able object. No need to wrap it again

src/extension/logger.ts
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/**
+ * Shows the output channel in the VS Code Output panel.
+ * This is useful for displaying error messages with a link to open the output.
+ */
+export function showOutputChannel(): void {
+  if (channel) {
+    channel.show(true);
+  } else if (outputChannelName) {
```
cbandera: We don't need the `else` branch here. Creating a channel should be the pure responsibility of `registerLogger`

src/test-explorer/lcov_parser.ts
```diff
@@ -123,7 +123,9 @@ async function demangleNameUsingFilter(
   mangled: string,
 ): Promise<string | undefined> {
   if (execPath === null) return undefined;
-  const unmangled = (await execFile(execPath, [mangled])).stdout.trim();
+  // c++filt needs the -n flag to properly demangle names that start with underscore
```
cbandera, line 126: This change seems unrelated. Might this also be something mac-related? The tests have worked as-is for me.

test/logger.test.ts
```diff
@@ -0,0 +1,573 @@
...
+    );
+    assert.ok(debugCall, "logVerbose should call debug");
+    assert.strictEqual(
+      debugCall.message,
+      "verbose message",
+      "logVerbose should format message correctly",
+    );
+  });
+
+  it("handles dispose correctly", () => {
```
cbandera, line 219: Let's also assert that the global vars are set back to undefined

src/extension/logger.ts
```diff
@@ -0,0 +1,275 @@
...
+  ...args: Arguments
+): void {
+  if (!channel) {
+    return;
+  }
+
+  // Log the detailed message to the output channel
+  let detailedMessage = message;
+  if (args.length > 0) {
+    detailedMessage += " Details:\n" + util.format(...args);
```
cbandera, line 89:
```suggestion
    detailedMessage += ":\n" + util.format(...args);
```
I feel the additional Details word is just confusing here:
> [debug] Running Bazel query Details:

src/extension/bazel_wrapper_commands.ts
```diff
@@ -431,7 +429,7 @@ async function bazelGoToLabel(target_info?: blaze_query.ITarget | undefined) {
 function copyLabelToClipboard(label: string): void {
   vscode.env.clipboard.writeText(label);
 
-  vscode.window.showInformationMessage(`Copied to clipboard: ${label}`);
+  logInfo(`Copied to clipboard: ${label}`, true);
```
cbandera: This message does also show the `Details` button, but there is really not any other info to see in the logging channel. 
I wonder whether we show introduce another flag to configure whether to show the button or not.

cbandera (commented):
Thank you for this great contribution! It's a feature that has already been requested several times.
I have done a first review and left you some findings.
In addition, I would like to ask you to consider introducing a new VSCode Command  `Bazel: Show output`. This way a user can always quickly bring up the output channel to view what's currently going on. An example implementation can be found [here](https://github.com/bazel-contrib/vscode-bazel/pull/496/changes/84fa7c4d1c1685d0c12d8011b73f8505b091bf6f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)

src/extension/bazel_wrapper_commands.ts
markkohdev: That makes sense 👍 
I think that if there's not really any details to show then we can just show the message without emitting a log.  Rather than trying to propagate something like `showDetailsButton: bool` all the way through the log methods I'll just add the extra param with a default to the `showUserMessage` function.

src/extension/logger.ts
markkohdev: Fair enough, I was trying to make this somewhat generic so that we could potentially add additional output channels (while still maintaining a global default) in the future if the extension scope grew (similar to having multiple logger classes in large applications), but in honesty it was a sort of half-baked effort that I didn't follow-through with since I made all of the logger methods reference the global default haha 😅  
I think it makes sense to be opinionated here for now.  If the need truly arises in the future to create a multi-channel logging system then we can address it then.

src/extension/logger.ts
markkohdev: I found that it helped for some messages, but I don't have a strong attachment to it so I can remove

src/extension/logger.ts
markkohdev: Yep, makes sense with the hard-coded channel.

src/test-explorer/lcov_parser.ts
markkohdev: yeah, I think this may be specifically due to the c++filt and rustfilt implementations for MacOS.  I take it you're not on a mac?

src/extension/logger.ts
markkohdev: The disposable wrapper was so that we could clear the globals.  Any suggestion for how to do that without the wrapping?  (typescript is definitely not my strong suit)

markkohdev:
@cbandera thanks for the review!  I've addressed most of your comments but let me know if you want a different approach for the disposable.

src/extension/logger.ts
cbandera: NVM. I don't have a strong opinion on it either.

src/extension/logger.ts
cbandera: I see, I didn't see the connection. But that's fine then.

src/test-explorer/lcov_parser.ts
cbandera: Yes. I am on linux.

Issue #444, opened 2025-03-20
Starlark Debugger does not start


Dragnalith:
I am trying to break on some BUILD.bazel file.

On the "Bazel Targets" section, when right-clicking on a target and selecting "Build Target with Starlark Debugger" thee VS Code debug tab open but the debugger does not start. For a few seconds I can see the debugger buttons (pause, start, step in, step out, ...) being displayed then disappear. Nothing is displayed in the Debug Console, no terminal run any commands, nothing appears in any VS Code output.

I have the same experience if I manually create a launch.json with the "bazel-launch-build" and trying to debug using it.

My environment:

- Windows 11
- VS Code: 1.98.2
- bazelbuild.vscode-bazel: 0.11.0

Issue #349, opened 2024-02-28
How to add breakpoint when “Build Target with Starlark Debugger”


oscarthecat:
i can't click on left side of BUILD.bazel file(current project) or .bzl file(source file in external dir) to add a breakpoint (red point)   in vscode.

Build Target with Starlark Debugger runs ok, it just no stop because no breakpoints.

my vscode version 1.86.2
bazel plugin version 0.9.0
macos version 14.3.1 (23D60)
bazel 6.2.1

how can i add breakpoint? thanks

cameron-martin:
Adding breakpoints works okay for me, although I am running Linux and don't have access to a Mac machine. Do you have a reproducer for this?

Strum355:
Breakpoints dont appear to work in macros, from some quick experimentation it appears to only work inside rules

cbandera:
Docs and the startup sequence for debugging have been improved. During testing breakpoints were working fine for me, also in macros.
Feel free to reopen, if this issue still isn't resolved for you with the current version on main.

Issue #192, opened 2020-03-26
Feature request: Make debugger accessible through Run/Add Configuration...


torgil:
> Currently, the Starlark Debugger can be used by right-clicking a build target in the Bazel Build Targets view and selecting "Build Target with Starlark Debugger".

This above option is currently non-optimal or non-functional  for a few scenarios
* Ability to debug a failing query phase
* Multiple variants of arguments to the same target

A proper "launcher"  with settings that could be saved in json with the ability to customize bazel commandline should solve this problem.

cbandera:
The available entrypoints and hints at how to configure have been added to the README.
Feel free to reopen if you still consider this unresolved.

Issue #184, opened 2020-02-08
Need Tutorial about Using the Starlark Debugger


shizukanaskytree:
Question: How to step debug bazel script?
Similar question: [how to debug bazel skylark script step by step with a debugger](https://stackoverflow.com/questions/50014336/how-to-debug-bazel-skylark-script-step-by-step-with-a-debugger)

According to [Using the Starlark Debugger](https://marketplace.visualstudio.com/items?itemName=BazelBuild.vscode-bazel&ssr=false#overview), I installed vscode-bazel on my Mac. 

> Currently, the Starlark Debugger can be used by right-clicking a build target in the Bazel Build Targets view and selecting "Build Target with Starlark Debugger". This will start the build inside the Visual Studio Code debugger (output will be redirected to the Debug Console pane) and it will pause on any breakpoints hit during execution.

> When a Bazel thread is paused, you can step through Starlark code, add watch expressions, and execute arbitrary statements by typing them in the input area of the Debug Console.

> Clicking the "Stop" button in the debugger will kill the Bazel process being debugger, allowing you to halt the current build. The Bazel server, however, will continue running.

By using a toy example,  code from[ Bazel tutorial example](https://docs.bazel.build/versions/2.0.0/tutorial/cpp.html)

```
(base) ~/Documents/02_prj/bazel_tutorial/04_bzl_debug$ls -1
README
WORKSPACE
bazel-04_bzl_debug@
bazel-bin@
bazel-out@
bazel-testlogs@
src/
```

```
(base) ~/Documents/02_prj/bazel_tutorial/04_bzl_debug/src$ls -1
BUILD
bzl.bzl
greeting.cc
greeting.h
hello.cc
```

I have no idea how to launch Starlark Debugger. Please provide a step-by-step tutorial, especially  screenshots of vscode.

marengaz:
don't have much experience with it, but this is the _Bazel Build Targets view_

![Screenshot 2020-05-13 at 18 33 17](https://user-images.githubusercontent.com/6930705/81845289-4b326800-9548-11ea-887d-93ab926735f2.png)

PR #558, opened 2026-02-01
feat(debug): improve Starlark debugger experience

cbandera:
- Automatically show debug view when debugging starts
- Force Starlark code reanalysis with --notrack_incremental_state flag
- Update documentation to improve usage instructions and clarify intent/scope of the debugger
- Update dependencies to latest version
- Fix an issue in buildifer which led to a crash when debugging into a external .bzl file

Closes: #444, #349, #192, #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Bazel plugin for Visual Studio Code
 
-[![Build Status](https://badge.buildkite.com/ecab448484315779ec28a95f8501b7f77a9a2abfc787037d5e.svg?branch=master)](https://buildkite.com/bazel/vscode-bazel-vs-bazel)
+[![Build Status](https://github.com/bazel-contrib/vscode-bazel/actions/workflows/build-release.yml/badge.svg)](https://github.com/bazel-contrib/vscode-bazel/actions/workflows/build-release.yml)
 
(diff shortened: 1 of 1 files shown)


src/extension/bazel_wrapper_commands.ts
```diff
@@ -154,6 +155,11 @@ async function bazelBuildTargetWithDebugging(
     request: "launch",
     type: "bazel-launch-build",
   });
+
+  // Automatically show the debug view when debugging starts
+  if (debugStarted) {
```
cameron-martin, line 160: I'm slightly skeptical of this change. Is this a pattern that other debugger extensions use? It seems like automatically switching tabs could be annoying if the user doesn't want to interact with this panel.

cbandera: I believe this is the intended behavior. 
It is also how Vscode behaves if you start a debugging session via F5. 

When the user executes this code path he selected "debug starlark code" from one of the entry points. As such we know the intention of the user is to debug, but for debugging you need the side pane.

cameron-martin: Okay makes sense

cameron-martin (approved):

PR #561, opened 2026-02-04
build: Bundle extension with esbuild

### LINKED ISSUE #538: Bundle extension during packaging step

cbandera:
Improves packages size from 5MB to 270kB
Improves activation time from 260ms to 85ms
Speeds up build significantly.

Closes #538

Closes #421 - Changes are in large parts taken from this PR that has stalled. Credits go to @cwahbong


package.json
```diff
@@ -495,17 +495,24 @@
...
-        "update-snapshot": "./scripts/test.sh -u",
-        "vscode:prepublish": "./scripts/build.sh",
-        "watch": "./scripts/build.sh -watch"
+        "lint-check": "eslint .",
+        "lint-fix": "eslint . --fix",
+        "build:debugadapter": "esbuild src/debug-adapter/client.ts --bundle --outfile=dist/debug-adapter-client.js --external:vscode --format=cjs --platform=node --minify",
+        "build:extension": "esbuild src/extension/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node --minify",
+        "build:tests": "esbuild test/**/*.test.ts --bundle --outdir=dist/test --external:vscode --format=cjs --platform=node --sourcemap",
+        "build": "npm run build:debugadapter && npm run build:extension",
+        "prebuild": "./scripts/prepare.sh",
```
cbandera, line 514: I am aware that the introduction of a prepare script has been rejected before.  
See https://github.com/bazel-contrib/vscode-bazel/pull/553#discussion_r2725874448

Note however that this PR goes a step further:
- My goal is to remove as many indirections as possible, modelling the build steps directly in the package Jason without calling scripts. 
- the npm run scripts are carefully crafted to assure that all necessary steps are always executed automatically. See pretest and pre build steps.

cbandera: Maybe we should rename this script into create protos and move the grammar step out of it

cbandera:
@cameron-martin , @vogelsgesang Can you have a look at the PR?

src/extension/extension.ts
```diff
@@ -35,7 +35,23 @@ import { activateWrapperCommands } from "./bazel_wrapper_commands";
 import { registerLogger, logInfo, logError, showOutputChannel } from "./logger";
 
 // Global reference to the workspace tree provider for testing
-export let _workspaceTreeProvider: BazelWorkspaceTreeProvider;
+declare global {
+  var bazelWorkspaceTreeProvider: BazelWorkspaceTreeProvider | undefined;
```
cameron-martin, line 39: How come we have this type definition but cast to `any` below?

package.json
cameron-martin: This is fine with me, as long as the whole build can be done in one command

cameron-martin (commented):
Mostly looks good to me, just one comment

src/extension/extension.ts
cbandera: Good catch. I believe the answer is: That is AI slop leftovers.
Removed them now.

cameron-martin (approved):

PR #564, opened 2026-02-09
feat: add `npm run try` for interactive extension testing

markkohdev:
## Description
Sorry @cbandera, here's another quick one for ya 😅 This was helpful when I was debugging some of the logging and querying changes.

- Adds a new `npm run try` script that packages the extension, downloads a local VS Code instance, installs the `.vsix`, and opens a disposable copy of the test Bazel workspace for interactive experimentation.
- The launched VS Code uses isolated user-data and extensions directories under `.vscode-test/`, so it doesn't affect the developer's normal VS Code installation.
- The test workspace is copied to `.vscode-test/try-workspace/` (skipping Bazel output symlinks) so edits made during experimentation don't modify source-controlled files.
- Documents the new workflow in CONTRIBUTING.md.

## Related Issue
N/A

## Testing
- [x] Run `npm run try` and verify VS Code launches with the extension active and the test workspace open.
- [x] Make edits in the launched VS Code and confirm `git status` shows no changes to `test/bazel_workspace/`.
- [x] Run `npm run clean` and confirm `.vscode-test/` is removed.
- [x] Run `npm run try` a second time and verify the workspace copy is refreshed (previous edits are gone).

## Checklist

### DIFF
--- a/scripts/try-extension.js
+++ b/scripts/try-extension.js
@@ -1,0 +1,151 @@
+#!/usr/bin/env node
+
+// Copyright 2026 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This script packages the extension, downloads a local VS Code instance (or
+// reuses a cached one), installs the packaged .vsix, and opens the test
+// workspace so you can experiment interactively.
+//
+// Usage:
+//   npm run try
+//
+// The VS Code instance uses isolated user-data and extensions directories
+// under .vscode-test/ so it won't affect your normal VS Code installation.
+
+const {
+  downloadAndUnzipVSCode,
+  resolveCliPathFromVSCodeExecutablePath,
+} = require("@vscode/test-electron");
+const { execSync, execFileSync, spawn } = require("child_process");
+const path = require("path");
+const fs = require("fs");
+
+/**
+1 * Recursively copy a directory, skipping symlinks. The Bazel convenience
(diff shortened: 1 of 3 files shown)


scripts/try-extension.js
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env node
+
+// Copyright 2026 The Bazel Authors. All rights reserved.
```
cbandera, line 3: BTW @cameron-martin do you know what the right copyright header is since the project moved to the linux foundation? Do we use copyright/license headers at all anymore?

scripts/try-extension.js
```diff
@@ -0,0 +1,151 @@
...
+  );
+
+  // Step 4: Copy the test workspace to a disposable location so edits
+  // made during experimentation don't touch the source-controlled original.
+  const srcWorkspace = path.join(projectRoot, "test", "bazel_workspace");
+  const workspaceCopy = path.join(projectRoot, ".vscode-test", "try-workspace");
+
+  // Always start from a fresh copy
+  if (fs.existsSync(workspaceCopy)) {
+    fs.rmSync(workspaceCopy, { recursive: true, force: true });
```
cbandera, line 116: Do we need to / should we cleanup the `userDataDir` and `extensionsDir` as well?

cbandera (commented):
I was inclined to say that we need another launch entrypoint. But given your explanation I do see that it would be useful (especially for PRs like #561 ).
And I think it could even be integrated with the `F5` entrypoints. E.g. via:
```json
        {
            "name": "Launch Extension (Isolated Test)",
            "type": "node",
            "request": "launch",
            "program": "${workspaceFolder}/scripts/try-extension.js",
            "console": "integratedTerminal",
            "env": {
                "NODE_ENV": "development"
            }
        }
```

I am having a hard time to decide what the best name would be. `try` seems to not really capture the whole nature of it.
Some ideas are: `sandboxed-deploy`, `isolated-launch`, `packaged-preview`... but I haven't found a favorite. Does any of these resonate with you?

Finally: It doesn't work on WSL which means I can't test it at the moment. The reason is that on WSL you typically start the GUI instance installed on windows and simply connect to your WSL instance through the "remote" extensions of VSCode. 
I will try to get a setup on linux again though to test this.

cbandera:
Tested it on native linux now and vscode doesn't start up either... :(

scripts/try-extension.js
cameron-martin: No idea. Maybe ask in the bazel slack?

PR #568, opened 2026-02-21
docs(contributing): add code structure guidelines and PR title validation

cbandera:
## Description

docs(contributing): add code structure guidelines and commit validation

## Related Issue

-

## Testing

New workflow step will be tested on this PR

## Checklist

### DIFF
--- a/.github/scripts/enforce-conventional-commit-message.sh
+++ b/.github/scripts/enforce-conventional-commit-message.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# Script to check if PR title follows the Conventional Commit format
+# Usage: .github/scripts/enforce-conventional-commit-message.sh [pr_title]
+
+set -euo pipefail
+
+# Conventional commit pattern - supports breaking changes (!), case-insensitive types, and custom types
+PATTERN='^[a-zA-Z]+(\(.+\))?!?: .+$'
+
+# Get PR title from argument
+PR_TITLE_TO_CHECK="$1"
+
+if [[ ! "$PR_TITLE_TO_CHECK" =~ $PATTERN ]]; then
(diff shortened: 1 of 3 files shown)


cameron-martin:
> I prefer to have squash commits on master. This keeps the history cleaner.
> 
> But as for the enforcement, I opted for checking both: commits and title. This might be annoying but helps to bring discipline into crafting the commits.
> 
> The only downside that I see with checking the title is that the workflow isn't retriggered automatically after editing it.

Either way they would be squashed, but I'm referring to how then default commit message is derived. Currently it's derived from the commits, rather than the PR title & description. 

I find the current option awkward to use if you want to avoid rewriting history on your branch, because you end up with a load of commits titled things like "address review comments", "fix linting", etc. Enforcing conventional commits on all commit messages is also awkward for this case, and is presumably designed for a workflow where the PR author rewrites history on their branch.

cbandera:
Thanks for clarifying again @cameron-martin . Yes I agree that using the PR title and description for the merge commit would best document changesets. Also because that is much easier to review and to directly edit as a maintainer.
At the same time it allows for the flexibility to quickly push a "fix typo" or "address review findings" commit on a existing PR.

I have adapted the PR to reflect this approach.

cameron-martin:
> Thanks for clarifying again @cameron-martin . Yes I agree that using the PR title and description for the merge commit would best document changesets. Also because that is much easier to review and to directly edit as a maintainer. At the same time it allows for the flexibility to quickly push a "fix typo" or "address review findings" commit on a existing PR.
> 
> I have adapted the PR to reflect this approach.

Okay thanks for confirming. I'll try find out how to change the GitHub setting then we can get this merged.

Issue #598, opened 2026-04-07
Add bazel.completion.queryScope to scope completion target query


runchen0919:
## Summary

Add a new `bazel.completion.queryScope` configuration option that allows users to limit the Bazel query used by the completion provider to specific package patterns, instead of always querying the entire workspace.

## Motivation

In large monorepos, the completion provider currently runs `kind('.* rule', ...)` which queries **all** targets in the workspace. This can be slow and return irrelevant results when a user only cares about a subset of packages.

## Proposed Change

Add a new setting `bazel.completion.queryScope` (type: `string[]`, default: `[]`) under the `bazel.completion` namespace.

- When **empty** (default), the existing behavior is preserved — the entire workspace is queried (`kind('.* rule', ...)`)
- When **non-empty**, the query is scoped to the specified patterns via union, e.g. `["//src/...", "//lib/..."]` → `kind('.* rule', //src/... + //lib/...)`

### Example configuration

```json
"bazel.completion.queryScope": ["//src/...", "//lib/..."]
```

## Difference from `bazel.commandLine.queryExpression`

| Setting | Type | Controls |
|---|---|---|
| `bazel.commandLine.queryExpression` | `string` | Workspace tree view & quick picker |
| `bazel.completion.queryScope` | `string[]` | BUILD file target auto-completion |

These two settings are independent and control different features.

cbandera:
Thanks for opening this issue. Since you already opened a PR, let's move the discussion of this issue there: https://github.com/bazel-contrib/vscode-bazel/pull/599#issuecomment-4208634482

PR #599, opened 2026-04-07
feat(completion): add queryScope config to scope completion target query

### LINKED ISSUE #123: Format deletes BUILD file if invalid
I have a problem where formatting of BUILD files using buildifier completely deletes the entire contents of the BUILD file. This is specifically when the syntax of the build file is incorrect. 

vscode-bazel version: 0.2.0
buildifier version: latest (git commit baa9c57c396019f0b7272b95d66f8e2681ed7d2a)

Steps to reproduce; 
1. Create invalid BUILD file 
2. Right click in the text editor window, then choose "Format document" (ctrl+shift+i)

runchen0919:
## Description

Add `bazel.completion.queryScope` setting (string array) that limits the Bazel query used by the completion provider to specific package patterns (e.g. `["//src/...", "//lib/..."]`), instead of always querying the entire workspace.

## Related Issue

https://github.com/bazel-contrib/vscode-bazel/issues/598

## Testing

- [x] Tests added/updated

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -197,6 +197,14 @@
                     "default": "...:*",
                     "description": "A [query language expression](https://bazel.build/query/language) which determines the packages displayed in the workspace tree and quick picker. The default inspects the entire workspace, but you could narrow it. For example: `//part/you/want/...:*`"
                 },
+                "bazel.completion.queryScope": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "default": [],
+                    "markdownDescription": "A list of Bazel package patterns to scope the completion provider's target query. When empty (default), the entire workspace (`...`) is queried. Specify patterns like `//src/...`, `//lib/...` to limit the query to specific directories. For example: `[\"//src/...\", \"//lib/...\"]`."
+                },
                 "bazel.lsp.command": {
                     "type": "string",
                     "default": "",
--- a/src/completion-provider/bazel_completion_provider.ts
+++ b/src/completion-provider/bazel_completion_provider.ts
@@ -18,6 +18,7 @@ import {
   getPackageLabelForBuildFile,
   queryQuickPickTargets,
 } from "../bazel";
+import { getCompletionQueryScope } from "../extension/configuration";
 
 function insertCompletionItemIfUnique(
   options: vscode.CompletionItem[],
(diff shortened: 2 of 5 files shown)


cbandera:
Thank you very much for your first contribution @runchen0919 ! 
The implementation looks clean and minimal. It's also nice to see that you added tests right away.

Conceptually though, I am sorry to say that I am not really in favor of this change. 
One reason is that we see a lot of requests to add new config options for specific needs - this has the potential downside of cluttering the extensions settings and to make it very confusing for future users to configure the extension properly. In order to better judge on the tradeoffs, could you maybe motivate why this change is necessary for your project setup? Is it the overhead of running full repository queries or do you see issues with the completion suggestions themselves?

The second reason is, that I am actually planning to remove the code that you have touched entirely, once the extension bundles a LSP directly. See also this comment: https://github.com/bazel-contrib/vscode-bazel/issues/1#issuecomment-3873266497

runchen0919:
> Thank you very much for your first contribution @runchen0919 ! The implementation looks clean and minimal. It's also nice to see that you added tests right away.
> 
> Conceptually though, I am sorry to say that I am not really in favor of this change. One reason is that we see a lot of requests to add new config options for specific needs - this has the potential downside of cluttering the extensions settings and to make it very confusing for future users to configure the extension properly. In order to better judge on the tradeoffs, could you maybe motivate why this change is necessary for your project setup? Is it the overhead of running full repository queries or do you see issues with the completion suggestions themselves?
> 
> The second reason is, that I am actually planning to remove the code that you have touched entirely, once the extension bundles a LSP directly. See also this comment: [#1 (comment)](https://github.com/bazel-contrib/vscode-bazel/issues/1#issuecomment-3873266497)

Hi, @cbandera Thank you for your reply. Let me explain the problem I encountered. My current project is a multi-language project managed by Bazel, including Python, Java, Go, and some code templates, etc. The code structure is as follows. In reality, as developers, we might only be concerned with the code in a specific directory, not scanning all targets in the root directory. There might be some problematic Bazel files in the code templates, causing the refresh function to encounter errors when scanning from the root directory.

```
| - code-template
    | - example.bzl   # error
| - src
    | - python
    |- go
    |- java
| - users
```

PR #608, opened 2026-04-13
fix: use Bazel workspace root as cwd instead of VS Code folder path

### LINKED ISSUE #607: BazelQuery uses VS Code workspace folder path instead of Bazel workspace root in multi-root workspaces
## Bug Description

When using VS Code in **multi-root workspace** mode, if a workspace folder is a subdirectory of the Bazel workspace (i.e., not the directory containing `MODULE.bazel`/`WORKSPACE`), the extension incorrectly uses that subdirectory path as the `cwd` when spawning Bazel processes. This causes `spawn` errors because Bazel commands are not executed from the correct workspace root.

## Steps to Reproduce

1. Create a Bazel project with the following structure:
   ```
   /project/
   ├── MODULE.bazel
   ├── src/
   │   └── BUILD
   └── tools/
       └── BUILD
   ```
2. Open VS Code with a multi-root workspace (`.code-workspace` file) that includes subdirectories as workspace folders:
   ```json
   {
     "folders": [
       { "path": "src" },
       { "path": "tools" }
     ]
   }
   ```
3. Observe that the Bazel Targets tree view fails to load with spawn errors.

## Root Cause

`BazelWorkspaceInfo` correctly stores both paths:
- `bazelWorkspacePath`: the Bazel workspace root (where `MODULE.bazel`/`WORKSPACE` lives)
- `workspaceFolder.uri.fsPath`: the VS Code workspace folder path (which can be a subdirectory)

However, 4 callsites pass `workspaceFolder.uri.fsPath` (the subdirectory) as `workingDirectory` to `BazelQuery`/`BazelCommand`, which then use it as `cwd` for `spawn`/`execFile`:

| File | Usage |
|------|-------|
| `bazel_workspace_folder_tree_item.ts` (×2) | `queryPackages` and `queryTargets` for tree view |
| `bazel_quickpick.ts` (×2) | `queryQuickPickTargets` and `queryQuickPickPackage` |

Other callsites (e.g., `bazel_package_tree_item.ts`, `tasks.ts`, `command_variables.ts`) already correctly use `bazelWorkspacePath`.

## Expected Behavior

All Bazel process spawning should use `bazelWorkspacePath` as the working directory, regardless of whether the VS Code workspace folder is a subdirectory.

## Environment

- VS Code with multi-root workspace
- Any Bazel version
- vscode-bazel extension

runchen0919:
## Summary

Fixes #607

In multi-root workspaces, VS Code workspace folders can be subdirectories of the Bazel workspace. Using these subdirectory paths as the working directory for spawning Bazel caused `spawn` errors.

## Changes

Changed 4 callsites to use `bazelWorkspacePath` (the directory containing `MODULE.bazel`/`WORKSPACE`) instead of `workspaceFolder.uri.fsPath`:

- **`bazel_workspace_folder_tree_item.ts`** (2 callsites): `queryPackages` and `queryTargets` for tree view rendering. Additionally, when the VS Code folder is a subdirectory, queries are automatically scoped to that subdirectory (`//<relative>/...:*` and `//<relative>:all`) for performance — avoiding a full workspace query.
- **`bazel_quickpick.ts`** (2 callsites): `queryQuickPickTargets` and `queryQuickPickPackage`.

## Why this is correct

Other callsites in the codebase already use `bazelWorkspacePath`:
- `bazel_package_tree_item.ts:60` — `this.workspaceInfo.bazelWorkspacePath` ✅
- `tasks.ts:164,281` — `workspaceInfo.bazelWorkspacePath` ✅
- `command_variables.ts:89,93,123` — `workspaceInfo.bazelWorkspacePath` ✅

This change makes the remaining 4 callsites consistent with the rest.

## Testing

- `tsc --noEmit` passes with zero errors
- Single-folder workspaces: behavior unchanged (`relativePath` is empty, original query expression used)
- Multi-root workspaces with subdirectory folders: Bazel now spawns from the correct root directory

### DIFF
--- a/src/bazel/bazel_query.ts
+++ b/src/bazel/bazel_query.ts
@@ -99,6 +99,7 @@ export class BazelQuery extends BazelCommand {
       .trim()
       .replace(/\r\n|\r/g, "\n")
       .split("\n")
+      .filter(Boolean)
       .sort();
     return result;
   }
--- a/src/bazel/bazel_query.ts
+++ b/src/bazel/bazel_query.ts
@@ -99,6 +99,9 @@ export class BazelQuery extends BazelCommand {
       .trim()
       .replace(/\r\n|\r/g, "\n")
       .split("\n")
+      // Remove empty strings caused by trailing newlines or empty output,
+      // e.g. "foo\nbar\n".split("\n") → ["foo","bar",""] — the trailing ""
+      // would otherwise pollute the package list.
       .filter(Boolean)
       .sort();
     return result;
--- a/src/workspace-tree/bazel_workspace_folder_tree_item.ts
+++ b/src/workspace-tree/bazel_workspace_folder_tree_item.ts
@@ -207,6 +207,11 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
       .replace(/\\/g, "/");
 
     const queryExpression = getQueryExpression();
+    // When the VS Code folder is a subdirectory of the Bazel workspace,
+    // `intersect` scopes the user's query to only packages under that
+    // subdirectory (//${relativePath}/...:*), and `except` removes the
+    // folder's own direct targets (//${relativePath}:*) so we only get
+    // sub-packages for the tree — direct targets are fetched separately below.
     const packageQuery = relativePath
       ? `((${queryExpression}) intersect (//${relativePath}/...:*)) except (//${relativePath}:*)`
       : queryExpression;
@@ -227,6 +232,9 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
 
     // Now collect any targets in the directory also (this can fail since
     // there might not be a BUILD files at this level (but down levels)).
+    // Intersect the user's query with the directory-level target query so
+    // only targets that satisfy both the user's filter and belong to this
+    // directory are shown (e.g. the user may restrict to certain rule kinds).
     const scopedTargetQuery = relativePath ? `//${relativePath}:all` : `:all`;
     const targetQuery = `(${queryExpression}) intersect (${scopedTargetQuery})`;
     const queryResult = await new BazelQuery(
--- a/src/workspace-tree/bazel_workspace_folder_tree_item.ts
+++ b/src/workspace-tree/bazel_workspace_folder_tree_item.ts
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import * as path from "path";
 import * as vscode from "vscode";
 import { BazelWorkspaceInfo, BazelQuery } from "../bazel";
 import {
@@ -199,11 +200,26 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
     if (!this.workspaceInfo || !this.workspaceInfo.workspaceFolder) {
       return Promise.resolve([]);
     }
-    const workspacePath = this.workspaceInfo.workspaceFolder.uri.fsPath;
+    // Always use the Bazel workspace root as cwd for spawning Bazel.
+    // In multi-root workspaces the VS Code workspace folder may be a
+    // subdirectory which would cause spawn errors when used as cwd.
+    const bazelWorkspacePath = this.workspaceInfo.bazelWorkspacePath;
+    const workspaceFolderPath = this.workspaceInfo.workspaceFolder.uri.fsPath;
+
+    // When the VS Code folder is a subdirectory of the Bazel workspace,
+    // scope queries to that subdirectory for performance.
+    const relativePath = path
+      .relative(bazelWorkspacePath, workspaceFolderPath)
+      .replace(/\\/g, "/");
+
+    const queryExpression = relativePath
+      ? `//${relativePath}/...:*`
+      : getQueryExpression();
+
     const packagePaths = await new BazelQuery(
       getBazelExecutablePath(),
-      workspacePath,
-    ).queryPackages(getQueryExpression());
+      bazelWorkspacePath,
+    ).queryPackages(queryExpression);
     const topLevelItems: BazelPackageTreeItem[] = [];
     this.buildPackageTree(
       packagePaths,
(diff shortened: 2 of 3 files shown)


src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -199,11 +200,26 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
...
+
+    // When the VS Code folder is a subdirectory of the Bazel workspace,
+    // scope queries to that subdirectory for performance.
+    const relativePath = path
+      .relative(bazelWorkspacePath, workspaceFolderPath)
+      .replace(/\\/g, "/");
+
+    const queryExpression = relativePath
+      ? `//${relativePath}/...:*`
+      : getQueryExpression();
```
cbandera: This change circumvents the user settings in `queryExpression`. 
Is it not enough to just change the `bazelWorkspacePath`?

I understand that you want to optimize for performance in the case that `queryExpression` defaults to `...:*`.
But if the user actually tries to use this setting, you will always overwrite it.

I vote for not messing with the query here and instead rely on the user optimizing this setting for his/her usecase locally.

src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -215,10 +231,11 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
 
     // Now collect any targets in the directory also (this can fail since
     // there might not be a BUILD files at this level (but down levels)).
+    const targetQuery = relativePath ? `//${relativePath}:all` : `:all`;
```
cbandera: Same as above. I vote against this optimization

cbandera (commented):
Thanks for opening this PR!
The core of this change looks correct, but I would prefer to merge without the runtime optimization, see findings below.

src/bazel/bazel_query.ts
```diff
@@ -99,6 +99,7 @@ export class BazelQuery extends BazelCommand {
       .trim()
       .replace(/\r\n|\r/g, "\n")
       .split("\n")
+      .filter(Boolean)
```
runchen0919, line 105: Problem: The queryPackages method splits Bazel query output by newline. When a query returns empty results, the output is an empty string "". .trim().split("\n") produces [""] (an array with one empty string element) instead of [].

Consequence: The empty string "" is passed to buildPackageTree as a package path, creating a BazelPackageTreeItem with packagePath="". Its getLabel() returns //${""} = //, producing a ghost node. Expanding it queries //:all, which displays the root BUILD file's contents.

Fix: .filter(Boolean) filters out empty strings so that empty results return [].

src/workspace-tree/bazel_workspace_folder_tree_item.ts
```diff
@@ -199,11 +200,22 @@ export class BazelWorkspaceFolderTreeItem implements IBazelTreeItem {
...
+    const packageQuery = relativePath
+      ? `((${queryExpression}) intersect (//${relativePath}/...:*)) except (//${relativePath}:*)`
+      : queryExpression;
+
     const packagePaths = await new BazelQuery(
       getBazelExecutablePath(),
-      workspacePath,
-    ).queryPackages(getQueryExpression());
+      bazelWorkspacePath,
+    ).queryPackages(packageQuery);
```
runchen0919, line 222: 1	(${queryExpression})	The user's configured query (e.g. kind('cc_library', ...:*))
2	intersect (//${relativePath}/...:*)	Intersect to scope results to the subdirectory
3	except (//${relativePath}:*)	Exclude the subdirectory's own package to avoid a duplicate entry with the BazelWorkspaceFolderTreeItem node

When at the workspace root (relativePath is empty), queryExpression is used directly without intersect.

src/workspace-tree/bazel_workspace_folder_tree_item.ts
runchen0919: Hi, @cbandera, Thank you very much for your reply. 

Good point — you're right that the previous version unconditionally overwrote queryExpression, which would silently ignore any user customization.

I've updated the approach: instead of replacing the user's queryExpression, I now use Bazel's intersect operator to combine it with the subdirectory scope:
((${queryExpression}) intersect (//${relativePath}/...:*)) except (//${relativePath}:*)

This way the user's queryExpression is always respected — intersect just narrows the results to the relevant subdirectory. If the user has set a custom expression to filter specific packages, it stays in effect; we only further scope it to the VS Code workspace folder's subtree.

As for "just change bazelWorkspacePath" — that alone fixes the spawn error (which was the original bug), but without any scoping the tree would display every package from the entire Bazel workspace, even if the VS Code folder only represents a subdirectory. That's both a performance and UX problem in large monorepos. The intersect approach gives us subdirectory scoping without sacrificing user control.

cbandera (approved):
Tested the changes locally and can confirm they are working.
Thank you for your explanations in the comments. That makes the diff better to understand. If you are very motivated, I would appreciate 2-3 comments in the source code explaining the `filter(boolean)`, the `intersect` and the `exclude` statement. That can help to avoid unintended regression in the future.

runchen0919:
> Tested the changes locally and can confirm they are working. Thank you for your explanations in the comments. That makes the diff better to understand. If you are very motivated, I would appreciate 2-3 comments in the source code explaining the `filter(boolean)`, the `intersect` and the `exclude` statement. That can help to avoid unintended regression in the future.

Thank you for your reply. I have added comments to the key changes.

runchen0919:
> > Tested the changes locally and can confirm they are working. Thank you for your explanations in the comments. That makes the diff better to understand. If you are very motivated, I would appreciate 2-3 comments in the source code explaining the `filter(boolean)`, the `intersect` and the `exclude` statement. That can help to avoid unintended regression in the future.
> 
> Hi, @cbandera Thank you for your reply. I have added comments to the key changes.

cbandera (approved):

PR #609, opened 2026-04-13
build: make prepare script cross-platform

### LINKED ISSUE #601: prepare.sh script not working under windows
Currently the extension is not buildable under windows, since the prebuild stage requires to call the prepare.sh script.

benjamin-maier96:
## Description

I added a prepare.js and a prepare.ps1 script to be able to build and debug the extension under windows.
In addition I fixed two minor bugs, where windows paths where not handled correctly.

## Related Issue

Closes #601 

## Testing

I was able to build and debug the extenstion under windows.

## Checklist


cbandera (commented):
Thank you for opening this PR @benjamin-maier96 !

While the change fixes your problem at hand, I fear it does introduce several new ones. E.g.
- it causes code duplication that will need to be maintained in the future
- it only adapt parts of the entry points to be windows-compatible (i.e. doesn't touch the `update_protos.sh` script)
- it suggests compatibility with windows, but that isn't enforced in CI

Since I am not working on windows natively, I am not really in favor of taking over the maintenance burden for another platform. But maybe we find a easy compromise: If you manage to rewrite the entrypoints in `node.js` we would have a cross-platform solution in a single implementation which doesn't cause maintenance overhead.

For the changes in the `src` folder I would suggest to open a separate PR to get these merged to `master` directly, as they are fixes for Windows at runtime as far as I understand.

benjamin-maier96:
Sounds reasonable. I only added the fixes in src because the unittests failed, but if we can live with failing unittests for windows for now, I will put them in a separate one.

benjamin-maier96:
I put all code in the prepare.js script. I tested it on a windows and linux machine. After this PR I will open another one for the runtime issues regarding the windows paths.

cameron-martin (approved):
LGTM, but will leave it for @cbandera to take a look too

benjamin-maier96:
I fixed the prettier warning.

cameron-martin:
This PR now has a load of unrelated changes

benjamin-maier96:
Oh, thanks for the catch. I fixed it. I hope that the format check is still passing, I tested it on my linux machine. On Windows it complains about every file, propably due to some file endings, but I will fix this later.

scripts/prepare.js
```diff
@@ -0,0 +1,56 @@
...
+// extension's build time.
+if (!fs.existsSync("src/protos/protos.js")) {
+  const protoFiles = fs
+    .readFileSync("src/protos/protos_list.txt", "utf8")
+    .split(/\r?\n/)
+    .map((line) => line.trim())
+    .filter((line) => line.length > 0)
+    .map((line) => `src/protos/${line}`);
+  execSync(
+    `npx pbjs -t static-module -o src/protos/protos.js ${protoFiles.join(" ")}`,
```
cbandera: Please update this line to contain the new `--path src/protos` argument that was added on the `master` branch.

cbandera (commented):
Changes look good. Please just resolve the merge conflict as hinted by the comment below.

Issue #490, opened 2025-12-11
Allow to selectively enable/disable features of this extension


cbandera:
We see several demands from users to selectively deactivate features of this extension. This is either because of performance reasons (long running queries in the background) or simply because features are not used and hence dependencies shouldn't be required (e.g. buildifier).

Features that qualify for deactivation:
- Target Tree in the explorer view
- CodeLens
- Bazel Commands requiring bazel (build/test/queries)
- Syntax Highlighting (? is this really something that you would want to turn off)
- Language Server
- Buildifier Linting

I assume there are some pitfalls to avoid here:
- Deactivating bazel commands will also render almost any other functionality useless
- We already have some settings that implicitly deactivate some features when the executable path is empty (language server). We might need to introduce a breaking change in order to provide a consistent user interface for configuration
- A lot of the setup code for the features is done during activation of the extension. It probably requires quite some refactoring to be able to deactivate features without requiring a reload of the complete IDE.

For reference, I saw demands of this here (and some of these could be closed if this issue is solved):
- https://github.com/bazel-contrib/vscode-bazel/pull/451
- https://github.com/bazel-contrib/vscode-bazel/issues/227
- https://github.com/bazel-contrib/vscode-bazel/issues/193
- https://github.com/bazel-contrib/vscode-bazel/issues/402
- https://github.com/bazel-contrib/vscode-bazel/issues/403
- https://github.com/bazel-contrib/vscode-bazel/issues/449
- https://github.com/bazel-contrib/vscode-bazel/issues/595

cbandera:
We should also clean up and cluster the config settings as soon as we have converted everything to extension features.

PR #613, opened 2026-04-19
feat: Allow to deactivate WorkspaceTree dynamically

cbandera:
## Description

This PR turns the `WorkspaceTree` into a `ExtensionFeature` which allows the user to selectively en-/disable the feature.

Additionally this changeset:
- Moves the setup code into the `WorkspaceTreeFeature`
- Renames the `bazel_workspace_tree_provider.ts` to `workspace_tree_provider.ts`
- Moves the check for Availability of a bazel workspace into the root of the extension, so that it is also available if the workspace tree has been deactivated.
- Extends the logger interface that is usable inside functions
- Adds debug logging to the WorkspaceTree Feature
- Activates the CodeLens Feature by default

## Related Issue

Related to #490 

## Testing

- [ ] Tests added/updated

## Checklist


cbandera:
@cameron-martin This is the next PR towards the goal described in #490 , following #496 .
I'd appreciate a review, so that I can continue with the next Features. (I don't want to work on parallel branches, and also want to avoid adding to many changes to a single PR).
The diff has gotten a little bigger because of the additional things that are mentioned in the PR description.

src/extension/extension.ts
```diff
@@ -122,27 +129,9 @@ export async function activate(context: vscode.ExtensionContext) {
...
     // Commands
     ...activateWrapperCommands(),
-
-    // Register command to manually refresh the tree view
-    vscode.commands.registerCommand("bazel.workspaceTree.refresh", () => {
-      _workspaceTreeProvider.refresh();
-    }),
-    vscode.commands.registerCommand("bazel.refreshBazelBuildTargets", () => {
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      completionItemProvider?.refresh();
```
cameron-martin, line 143: We're no longer refreshing the completion item provider in response to `bazel.refreshBazelBuildTargets`. Was this intentional? What are the implications here?

package.json
```diff
@@ -104,7 +104,8 @@
                 "icon": {
                     "dark": "./icons/refresh-dark.svg",
                     "light": "./icons/refresh-light.svg"
-                }
+                },
+                "when": "bazel.workspaceTree.enabled"
```
cameron-martin: Isn't this `bazel.feature.WorkspaceTree.enabled`?

cameron-martin: There's also a few views that need to be gated on this too, otherwise when you disable the feature you get an empty view that is broken.

cameron-martin (changes requested):

src/extension/extension.ts
cbandera: That's true and I must have overseen this.
Also I wasn't actually aware until now that the "Refresh Bazel Build Targets" command also had this side effect (I wonder whether our users knew?

The implications of removing this is, that at the moment there is no interactive way of updating the providers knowledge about symbols (except restarting VSCode..).

One solution would be to create a file watcher and to refresh the providers knowledge whenever a BUILD file has changed. That would be the most reliable approach because users are guaranteed to always have the most recent completion suggestions. But on the downside, that could cause a lot of background queries.

I will go with this approach anyway, as I also plan to provide a feature switch for this language server functionality in one of the next PRs anyway. That would provide a way to opt-out of the queries again.

package.json
cbandera: Thank you, this is yet another good catch. 
I have corrected the `when` statement on both the command and the views and tested it locally.

cameron-martin (approved):

Issue #428, opened 2025-01-03
Extension cannot find bazel root without a WORKSPACE file


leryss:
In pure bzlmod projects where there is no WORKSPACE file, the extension cannot find the root of the project and all paths become absolute as seen in the screenshot below. Is this intended?

![image](https://github.com/user-attachments/assets/b8cc20b5-d819-4c68-a4bc-8c7992886857)

cameron-martin:
A reproducer for this would be appreciated.

leryss:
remove the `WORKSPACE` file and restart vscode
go to any target and the gray path added by the extension will become absolute

cameron-martin:
Unfortunately I cannot reproduce this. If you could give me a repository to reproduce this, plus tell me the version of the extension you are using, that would be helpful.

PR #625, opened 2026-04-26
Feat/manual workspace path

### LINKED ISSUE #472: bazel 8 + bzlmod and child modules resolution
# VS Code Bazel extension does not respect root toolchains / .bazelrc in multi-module bzlmod monorepos

## Problem description

When using **Bazel with bzlmod** in a monorepo, each app or library is its own Bazel **module** (`apps/appX/MODULE.bazel`, `modules/shared/MODULE.bazel`, etc.).  
The monorepo root also has a `MODULE.bazel` where all **toolchains** and the main **.bazelrc** are centralized.

### Expected behavior
- The VS Code Bazel extension should treat the **monorepo root** as the single Bazel workspace root.  
- All build/test/run actions (CodeLens, Test UI, right-click commands) should run Bazel from the **repo root**, so that:
  - the root `.bazelrc` is applied,
  - toolchains registered in the root `MODULE.bazel` are available,
  - cross-module deps work (e.g. `@other_module//path:target`),
  - labels are consistent (`//modules/transport/...:foo`) rather than relative to a leaf folder.

### Actual behavior
- The extension currently picks the **nearest `MODULE.bazel` or WORKSPACE file** to the open file and treats that as the workspace root.  
- If I open `apps/app-go/src/main.go` and run “Bazel: Build Target”, the extension runs Bazel in `apps/app-go/`.  
  - Labels resolve relative to that folder (e.g. `//proto/...` instead of `//modules/transport/proto/...`).  
  - Toolchains defined in the root `MODULE.bazel` are not registered.  
  - The root `.bazelrc` is ignored.  
- This makes cross-module builds/tests fail when triggered from VS Code, even though they succeed when run manually from the monorepo root.

### Why this matters
- With **bzlmod**, the recommended monorepo structure is *multiple modules with a root module* that wires toolchains and overrides.  
- IDE users (VS Code) expect Bazel commands to “just work” without having to `cd` into the root manually.  
- As of today, developers must either:
  - open only the repo root in VS Code (which breaks per-app workspace setups), or  
  - duplicate toolchain registration and `.bazelrc` in every leaf module (which defeats the purpose of centralization).

## Steps to reproduce
1. Create a monorepo with:
MODULE.bazel # root, toolchains + local_path_override
apps/app-go/MODULE.bazel # leaf module
2. Open `apps/app-go/` in VS Code.  
3. Run “Bazel: Build Target” on a file.  
4. Observe: Bazel runs from `apps/app-go/` instead of the monorepo root, toolchains are missing, and labels resolve incorrectly.

## Desired solution
- Provide a VS Code setting to **force Bazel commands to run from the top-level workspace root** (not the nearest).  
- Example: `"bazel.workspaceRootMode": "repo-root-only"`.  
- Or auto-detect the “primary” root (the one containing `.git/` and `MODULE.bazel`) instead of treating every nested module as independent.  
- At minimum, document that opening a leaf module as the VS Code workspace will break root toolchains/config.

## Environment
- Bazel: 8.x with bzlmod enabled
- VS Code: latest
- Extension: vscode-bazel latest
- OS: Any

### LINKED ISSUE #621: getBazelWorkspaceFolder for multiple MODULE.bazel files
The comment here doesn't seem to match the code https://github.com/bazel-contrib/vscode-bazel/blob/bafb21a1de6fb65eb63b480fb175d900c634728b/src/bazel/bazel_utils.ts#L143-L162.

Actually, we have a use case where we would need the behavior of the comment. We have the following folder structure

root
-- MODULE.bazel (with a local_repository_override for the internal_dependency)
-- internal_dependency
---- MODULE.bazel (with missing local_repository_override for other internal dependencies)

The folder structure comes from the fact that we do not yet have a monorepo, but clone multiple different internal repos in one root folder. We then use the root MODULE.bazel together with local_repository_override to target the different repositories and make them available to bazel. The individual MODULE.bazel files in the internal depedencies then do not have those local_repository_overrides.

This works quite well for us, but the extension is not able to deduce the root MODULE.bazel file and therefore the queries are failing, because the extension performs then inside of the internal_dependency folder and not the root folder.

If the intended behavior is to use the nearest MODULE.bazel file, a solution to our problem would be to introduce a WORKSPACE_FOLDER parameter, that allows us to pin the root MODULE.bazel file.

Zsi-r:
## Summary

Add a new configuration option `bazel.workspacePath` that allows users to manually specify the Bazel workspace directory path.

## Problem

Fixes #621
Fixes #472

When working with multiple `MODULE.bazel` files in nested directories (e.g., monorepo setups with `local_path_override`), the extension picks the nearest workspace file instead of the intended root workspace. This causes queries to fail because they run in the wrong directory.

**This is especially useful when opening a subdirectory of the Bazel workspace in VS Code**, where the root `MODULE.bazel`/`WORKSPACE` file is located in a parent directory outside the opened folder.

Related: #428

## Solution

Added a new configuration option `bazel.workspacePath` that:

- Accepts an absolute path or a path relative to the VS Code workspace folder
- Takes precedence over auto-detection when set
- Falls back to the existing auto-detection behavior when not configured

## Usage

```json
{
  "bazel.workspacePath": "/absolute/path/to/workspace"
}

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,6 +165,12 @@
                     "default": false,
                     "markdownDescription": "Whether to add a CodeLens to `BUILD`/`BUILD.bazel` files to provide actions while browsing the file."
                 },
+                "bazel.workspacePath": {
+                    "type": "string",
+                    "default": "",
+                    "markdownDescription": "Manually specify the path to the Bazel workspace directory. This can be an absolute path, or a relative path from the VS Code workspace folder. If set, this path will be used instead of auto-detecting the workspace by searching for `MODULE.bazel`, `REPO.bazel`, `WORKSPACE.bazel`, or `WORKSPACE` files.",
+                    "scope": "resource"
+                },
                 "bazel.pathsToIgnore": {
                     "type": "array",
                     "items": {
--- a/src/bazel/bazel_utils.ts
+++ b/src/bazel/bazel_utils.ts
@@ -14,8 +14,9 @@
 
 import * as fs from "fs";
 import * as path from "path";
+import * as vscode from "vscode";
 import { blaze_query } from "../protos";
-import { getPathsToIgnore } from "../extension/configuration";
+import { getPathsToIgnore, getWorkspacePath } from "../extension/configuration";
 import { logError } from "../extension/logger";
 import { BazelQuery } from "./bazel_query";
 
@@ -139,8 +140,70 @@ function findAncestorFile(
 }
 
 /**
+ * Resolves the manually configured workspace path to an absolute path.
+ *
+ * @param fsPath The path to a file, used to determine the VS Code workspace folder.
+ * @returns The resolved absolute workspace path, or undefined if not configured or invalid.
+ */
+function resolveConfiguredWorkspacePath(fsPath: string): string | undefined {
+  const uri = vscode.Uri.file(fsPath);
+  const workspaceFolder = vscode.workspace.getWorkspaceFolder(uri);
+  const configuredPath = getWorkspacePath(uri);
+
+  if (!configuredPath) {
+    return undefined;
+  }
+
+  let resolvedPath: string;
+
+  // Check if it's an absolute path
+  if (path.isAbsolute(configuredPath)) {
+    resolvedPath = configuredPath;
+  } else if (workspaceFolder) {
+    // Resolve relative path from VS Code workspace folder
+    resolvedPath = path.join(workspaceFolder.uri.fsPath, configuredPath);
+  } else {
+    // No workspace folder, try to resolve from the file's directory
+    resolvedPath = path.resolve(path.dirname(fsPath), configuredPath);
+  }
+
+  // Verify the path exists and contains a workspace file
+  try {
+    const stat = fs.statSync(resolvedPath);
+    if (stat.isDirectory()) {
+      // Check if any workspace file exists in the directory
+      const workspaceFiles = [
+        "MODULE.bazel",
+        "REPO.bazel",
+        "WORKSPACE.bazel",
+        "WORKSPACE",
+      ];
+      for (const file of workspaceFiles) {
+        try {
+          fs.accessSync(path.join(resolvedPath, file), fs.constants.F_OK);
+          return resolvedPath;
+        } catch {
+          // File not found, continue
+        }
+      }
+      // No workspace file found, but directory exists - still use it
+      // (user explicitly configured it)
+      return resolvedPath;
+    }
+  } catch {
+    // Path doesn't exist
+    return undefined;
+  }
+
+  return undefined;
+}
+
+/**
  * Search for the path to the directory that has the Bazel WORKSPACE file for
  * the given file.
+ *
+ * If a workspace path is manually configured via `bazel.workspacePath`, it will
+ * be used instead of auto-detection.
  *
  * If multiple directories along the path to the file have workspace files,
  * the lowest path is returned.
@@ -150,6 +213,13 @@ function findAncestorFile(
  * otherwise undefined.
  */
 export function getBazelWorkspaceFolder(fsPath: string): string | undefined {
+  // First, check if a workspace path is manually configured
+  const configuredWorkspace = resolveConfiguredWorkspacePath(fsPath);
+  if (configuredWorkspace) {
+    return configuredWorkspace;
+  }
+
+  // Fall back to auto-detection
   const workspaceFile = findAncestorFile(fsPath, [
     "MODULE.bazel",
     "REPO.bazel",
--- a/src/extension/configuration.ts
+++ b/src/extension/configuration.ts
@@ -57,6 +57,19 @@ export function getPathsToIgnore(): string[] {
   return getConfigurationWithDefault<string[]>("bazel", "pathsToIgnore");
 }
 
+/**
+ * Gets the manually specified Bazel workspace path from the workspace
+ * configuration.
+ *
+ * @param scopeUri Optional URI to scope the configuration lookup to a specific
+ * resource.
+ * @returns The manually specified workspace path, or an empty string if not set.
+ */
+export function getWorkspacePath(scopeUri?: vscode.Uri): string {
+  const config = vscode.workspace.getConfiguration("bazel", scopeUri);
+  return (config.get<string>("workspacePath") || "").trim();
+}
+
 export function getStartupOptions(): string[] {
   return getConfigurationWithDefault<string[]>(
     "bazel.commandLine",
--- a/test/bazel_utils.test.ts
+++ b/test/bazel_utils.test.ts
@@ -1,9 +1,11 @@
 import * as path from "path";
 import * as fs from "fs";
 import * as assert from "assert";
+import * as vscode from "vscode";
 import {
   getBuildFileLineWithSourceFilePath,
   getTargetNameAtBuildFileLocation,
+  getBazelWorkspaceFolder,
 } from "../src/bazel/bazel_utils";
 
 const workspacePath = path.join(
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

**Comment by cbandera** [comment]
We recently merged #608. Might this maybe already be enough to fix your problem? It at least should enable you to open a subdirectory of a Bazel workspace and have the queries run from the root of it. 

As I am generally against adding to many config options I would like to understand your usecase better. What is the motivation for opening just a subdirectory of a workspace instead of the true root?

**Comment by benjamin-maier96** [comment]
> We recently merged #608. Might this maybe already be enough to fix your problem? It at least should enable you to open a subdirectory of a Bazel workspace and have the queries run from the root of it.

I don't think it helps for the linked issues #621 and #472. The problem here is that we have multiple MODULE.bazel files and we run all queries from the nearest MODULE.bazel file. But there exists project setups, where only the root MODULE.bazel file contains all the resolved paths and therefore every query must be run from the root folder.

The PR you mentioned is only able to find the MODULE.bzl file if we have only a subfolder open.

**Comment by cameron-martin** [comment]
I do think this is a reasonable request, although I would be inclined to make it an array, like the `linkedProjects` config in rust-analyzer, and resolve the closest parent directory in this array from the current file.

There was a previous discussion on how to automatically fix the situation you're encountering in a [slack thread](https://bazelbuild.slack.com/archives/C014RARENH0/p1702897315236089?thread_ts=1702897315.236089&cid=C014RARENH0) a while ago, but doing this correctly is tricky so a config option sounds better here.

**Comment by benjamin-maier96** [comment]
I tested the changes on my windows machine and it picks up the correct path.

However, currently not all queries are executed from bazel root folder. For example the go to definition query uses the current document as working directory.

https://github.com/bazel-contrib/vscode-bazel/blob/759d20811c11bc6441279d9f542d9c3f034f93dc/src/definition/bazel_goto_definition_provider.ts#L80-L101
 
 I assume this is simply a bug and we should use `workspaceInfo.bazelWorksapcePath` as the working directory. With this change my issue #621 would be resolved. Besided that we might have more code lines where we would need to use the correct working directory.

**Comment by benjamin-maier96** [comment]
Any updates on this one?

**Comment by benjamin-maier96** [comment]
@cbandera @cameron-martin 
What is the current state of this pull-request? If we can agree on the extra parameter I would be willing to finish the necessary changes.

**Comment by Karnabanu** [comment]
Hi @cbandera @cameron-martin,

I would like to help move this feature forward. I tested #625 against a large Bazel workspace with the following layout:

```text
vscode-workspace/
└── bazel_root/
    └── MODULE.bazel
```

This case is not covered by #608: the VS Code workspace contains the Bazel root rather than being located beneath it.

While testing, I found a remaining issue. With a configuration such as:

```json
{
  "bazel.workspacePath": "bazel_root"
}
```

#625 successfully locates the workspace, but the workspace-tree logic calculates a relative path of `..` and generates an invalid `//../...` query.

I prepared a supplemental fix here:

https://github.com/Zsi-r/vscode-bazel/pull/1

It:

- handles root, descendant, containing-parent, and unrelated paths safely;
- prevents invalid `//../...` queries;
- fixes editor-to-tree path resolution;
- adds additional tests;
- passes type checking and 22 targeted tests;
- has been manually validated with the nested-root layout.

Could you confirm the preferred direction for `bazel.workspacePath`? I noted the earlier suggestion of supporting an array similar to rust-analyzer’s `linkedProjects`, and I am happy to adapt the implementation if that design is preferred.

If the current single-path design is acceptable and #625 is merged, I can immediately submit the supplemental fix directly to `bazel-contrib/vscode-bazel`. Alternatively, because the original branch has not been updated recently, I can submit a rebased replacement PR against the current `master` containing both changes while preserving the original author’s attribution.

Please let me know which approach would be easiest to review and merge.

src/bazel/bazel_utils.ts
```diff
@@ -138,10 +139,72 @@ function findAncestorFile(
...
+        try {
+          fs.accessSync(path.join(resolvedPath, file), fs.constants.F_OK);
+          return resolvedPath;
+        } catch {
+          // File not found, continue
+        }
+      }
+      // No workspace file found, but directory exists - still use it
+      // (user explicitly configured it)
+      return resolvedPath;
```
cbandera, line 191: Currently, whether or not `MODULE.bazel`/`WORKSPACE`/etc. is found in the configured directory, the function returns the directory anyway — the check is dead code that misleads readers (and gives users no signal when they've mistyped `bazel.workspacePath`). Either return `undefined` (and log an error) when no marker file is present, or drop the loop and document that any existing directory is accepted.
I would prefer to log an error, as this could likely be a mistake like a mistyped path and using it could cause confusing errors downstream from here

package.json
```diff
@@ -165,6 +165,12 @@
                     "default": false,
                     "markdownDescription": "Whether to add a CodeLens to `BUILD`/`BUILD.bazel` files to provide actions while browsing the file."
                 },
+                "bazel.workspacePath": {
+                    "type": "string",
+                    "default": "",
+                    "markdownDescription": "Manually specify the path to the Bazel workspace directory. This can be an absolute path, or a relative path from the VS Code workspace folder. If set, this path will be used instead of auto-detecting the workspace by searching for `MODULE.bazel`, `REPO.bazel`, `WORKSPACE.bazel`, or `WORKSPACE` files.",
```
cbandera, line 171: Can you state the "one root per VS Code folder" contract explicitly, and point at multi-root workspaces as the way to get more than one.

cbandera (commented):
Please excuse my long silence - I do hate it myself if Open Source projects stall. I did have a lot of other stuff lately and my gut feeling was that this is a bigger issue that requires more time to investigate thoroughly.

I did try to really understand this issue now and to correlate it with other issues that were reported in order to make a design decision that is not a local "balcony" or a workaround that will break again in a different location. You can read up on (or feed to your copilot) the full picture in #684 .

Long story short: Both your proposed changes are exactly right and are both required to solve the flaws in the current design. So best to merge them together into one.
In addition, the following points need addressing:

1. Fix `bazel_goto_definition_provider.ts` to use `workspaceInfo.bazelWorkspacePath` as the query cwd instead of `Utils.dirname(document.uri)` (line 80-82). This is the same class of bug #608 already fixed at 4 other call sites and Zsi-r#1 fixed at a 5th; goto-definition is the one remaining outlier (according to my research).
5. Verify against recent changes from #638:  Once `bazel.workspacePath` is set and `resolveConfiguredWorkspacePath`'s containment check from Zsi-r#1 is in place, `fromDocument` for a file inside a nested override'd module should resolve to the *pinned* root instead of nearest-wins, which should make it match the cache key populated via `fromWorkspaceFolder` for the same folder. Maybe even add a regression test in `test/completion.test.ts` for S5 with `bazel.workspacePath` set.
4. Tests: Is it possible to add an end-to-end-ish test (or at least a tree-provider unit test) that exercises S4 (see design decision) through the actual tree-building code path, not just `getBazelWorkspaceFolder()` in isolation — the current tests all passed while the real S4 bug (invalid `//../...` query) was only found by @Karnabanu though manuall testing.
2. (low priority) `bazel_availability.ts:25` only checks `workspaceFolders[0]` for a relative bazel executable path; doesn't consult `bazel.workspacePath`.
3. (could be a followup) Update the contributing docs to state the new design decision of explicitly only supporting a single root

Is this something either of you could do?

Karnabanu:
Hi @cbandera,

Thank you for the detailed review and for documenting the design decision in #684.

I have opened #687 as a consolidated draft replacement for this PR. It preserves @Zsi-r’s original commits and authorship, includes the nested workspace-root fixes, and addresses the five review items listed above.

I have also closed Zsi-r/vscode-bazel#1 so that the remaining review and discussion can stay in the upstream repository.

Could you please close this PR and continue the review and discussion on #687?

src/bazel/bazel_utils.ts
Karnabanu: Addressed in #687. A configured directory without a Bazel workspace marker now logs an error and returns `undefined`, allowing normal workspace auto-detection to take over. I also added a regression test for this case.

package.json
Karnabanu: Addressed in #687. The setting description and README now explicitly state that one Bazel root is supported per VS Code workspace folder and recommend a VS Code multi-root workspace for multiple independent roots.

cbandera:
Closing in favor of #687

Issue #402, opened 2024-06-21
make buildifier warning less aggressive


jab:
This extension's warning about no `buildifier` being found is overly eager, and presumes that the environment should have a single, global `buildifier` executable shared by multiple projects.

But developers often prefer to have project-local versions of their tool dependencies (and no global version) for better reproducibility, which is even facilitated for `buildifier` by https://github.com/keith/buildifier-prebuilt. Note that all projects created via `aspect init` use this setup (see https://github.com/aspect-build/aspect-workflows-template).

Such users can add something like the following to ensure that a `buildifier` binary is available when their projects are opened and that the Bazel extension is configured to use it:
.vscode/tasks.json:
```json
    {
      // Build `buildifier` on folderOpen to ensure that a buildifier binary exists which the
      // VSCode Bazel plugin can use.
      "label": "Build buildifier",
      "command": "bazel build @buildifier_prebuilt//:buildifier",
      "presentation": {
        "reveal": "never",
        "panel": "new"
      },
      "runOptions": {
        "runOn": "folderOpen",
      },
      "type": "shell",
    }
```
.vscode/settings.json:
```json
{
    "bazel.buildifierExecutable": "bazel-bin/external/buildifier_prebuilt~/buildifier/buildifier",
}
```
But even with this configuration, this extension still nags about no `buildifier` being found, because the check for it completes a split second before it's finished building, and it's never checked for again.

If the initial check fails, how about doing a retry or two with a sensible backoff?

cc @alexeagle

jab:
Aha, looks like the config above can be changed to just setting `bazel.buildifierExecutable` to `@buildifier_prebuilt//:buildifier` and then this extension will use bazel to build that target and use it, nice.

alexeagle:
Ooh, did you find that documented? If not I would like to re-open this to track the missing docs

cameron-martin:
Looks like this was added in https://github.com/bazelbuild/vscode-bazel/pull/350, but there doesn't seem to be docs for this. I'll re-open this to track docs support. A PR is very much welcome :)

jab:
Thanks for reopening this in support of making the user experience better here.

The fine print that is already in the Settings UI is indeed how I eventually discovered this:
<img width="1228" alt="Screenshot 2024-06-23 at 1 23 26 PM" src="https://github.com/bazelbuild/vscode-bazel/assets/64992/e1165221-7a76-4114-aa2c-c744538154f7">

[...]

PR #629, opened 2026-05-03
feat(Buildifier): Allow enabling /disabling buildifier functionality 

### LINKED ISSUE #193: don't throw when buildifier is not found
We'd like to just use this extension for code highlighting. So we don't need to have buildifier or bazel installed globally on our systems.
Possibly adding config to disable this behaviour would suffice

### LINKED ISSUE #449: Want a config to toggle buildifier lint diagnostics
The background is: many people are working with older bazel, so many new lint diagnostics are not helpful.

### LINKED ISSUE #620: Buildifier not found in path if installed via mise
If I install buildifier via https://mise.jdx.dev/, the VScode extension does not pick it up on the path.

Placing `echo "$(which buildifier)"` inside `.zprofile` correctly outputs `/Users/ivan/.local/share/mise/shims/buildifier` inside the vscode terminal, and it's available on the path.

I also have the following, which does not help:
```json
    "terminal.integrated.automationProfile.osx": {
        "path": "/usr/bin/zsh",
        "args": ["--login"]
    }
```

Installing it globally via mise does not help either.

Setting the path in settings to `/Users/ivan/.local/share/mise/shims/buildifier` or `/Users/ivan/.local/share/mise/installs/buildifier/8.5.1/buildifier` does not help either.

cbandera:
## Description

- Add BuildifierFeature class extending BaseExtensionFeature
- Add bazel.enableBuildifier configuration setting (default: true)
- Move buildifier diagnostics and formatting provider registration into feature
- Extract showBuildifierDownloadPrompt for async user notifications
- Remove duplicate buildifier availability checks from extension.ts
- Maintain backward compatibility with existing buildifier functionality
- Make buildifier feature independent of the presence of a bazel workspace (now uses absolute paths)

This change encapsulates buildifier support as a toggleable feature following the same patterns as CodeLensFeature and WorkspaceTreeFeature.

## Related Issue

Related to #490 
Closes #193
Related to #402
Closes #449
Closes #620 

## Testing

- [x] Tests added/updated

## Checklist

- [X] Commit messages follow [Conventional Commit](https://www.conventionalcommits.org/) conventions

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,6 +170,11 @@
                     "default": true,
                     "markdownDescription": "Whether to show Bazel packages and targets in a tree view in the Explorer panel. Deactivate to avoid running queries in the background."
                 },
+                "bazel.enableBuildifier": {
+                    "type": "boolean",
+                    "default": true,
+                    "markdownDescription": "Whether to enable Buildifier support for formatting and linting Bazel files. When enabled, provides document formatting and lint diagnostics for BUILD, WORKSPACE, and .bzl files."
+                },
                 "bazel.pathsToIgnore": {
                     "type": "array",
                     "items": {
--- a/src/buildifier/buildifier_feature.ts
+++ b/src/buildifier/buildifier_feature.ts
@@ -1,0 +1,83 @@
+// Copyright 2018 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as vscode from "vscode";
+
+import { BaseExtensionFeature } from "../extension/extension_feature";
+import {
+  checkBuildifierIsAvailable,
+  showBuildifierDownloadPrompt,
+} from "./buildifier_availability";
+import { BuildifierDiagnosticsManager } from "./buildifier_diagnostics_manager";
+import { BuildifierFormatProvider } from "./buildifier_format_provider";
+
+/**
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -170,6 +170,11 @@
                     "default": true,
                     "markdownDescription": "Whether to show Bazel packages and targets in a tree view in the Explorer panel. Deactivate to avoid running queries in the background."
                 },
+                "bazel.enableBuildifier": {
```
cbandera, line 173: One design question that we should maybe clarify @cameron-martin is whether we want to use the technical term `Buildifier` for this feature (which probably feels right for everyone familiar with the tool)
Or whether we want to follow a more use-case centric approach and name it `FormattingAndDiagnostics` or similar.

src/buildifier/buildifier_feature.ts
```diff
@@ -0,0 +1,85 @@
...
+export class BuildifierFeature extends BaseExtensionFeature {
+  private diagnosticsManager?: BuildifierDiagnosticsManager;
+
+  constructor(context: vscode.ExtensionContext) {
+    super("Buildifier", context);
+  }
+
+  enable(context: vscode.ExtensionContext): boolean {
+    // Precondition: buildifier executable available
+    if (!checkBuildifierIsAvailable()) {
```
cameron-martin: The `checkBuildifierIsAvailable` function is async so this condition will always evaluate to false (since a promise is truthy). Is it viable for this function to be async?

This would have been caught if we had the [no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises/) eslint option enabled. I'll put up a PR to enable this.

cameron-martin (changes requested):

package.json
cameron-martin: I think Buildifier seems like the most natural naming, but I could be skewed by the existing implementation in the extension.

src/buildifier/buildifier_feature.ts
cbandera: Thanks. I will need to think about the best design for this a little longer

cbandera: Made the `enable()` function async, which however also made it necessary to avoid race conditions with multiple config changes while a enable call is still running.

cameron-martin (approved):

PR #636, opened 2026-05-11
feat(LanguageSupport): Allow enabling/disabling language support functionality

cbandera:
## Description

- Create single LanguageSupportFeature that handles both external LSP and built-in implementations
- Automatically choose implementation based on bazel.lsp.command configuration (as it was)
- Refactor LSP client management to prevent duplicate output channels (when re-enabling)
- Ensure proper resource cleanup and disposable management
- Rename lsp folder to language_support

## Related Issue

Related to #490

## Testing

- [ ] Tests added/updated
- [x] Manual testing performed

## Checklist

- [x] Code follows the project's style guidelines (prettier/eslint)
- [x] Commit messages follow [Conventional Commit](https://www.conventionalcommits.org/) conventions
- [x] Tests pass locally (`npm run test`)

PR #638, opened 2026-05-14
fix: silence workspace error in non-Bazel repos

### LINKED ISSUE #446: "Failed to find a Bazel Workspace file" on non bazel repos that include BUILD files
I don't know if this is intended behavior, but for every repo I open that doesn't use bezel, but does contain `BUILD` files for a different build system, I get the following error:

`Failed to find a Bazel workspace`

I'm wondering if this is the intended behavior, or if there's a regression. I found [this bug](https://github.com/bazel-contrib/vscode-bazel/issues/314) from 2023 which seems to have been fixed.

This may also be related to https://github.com/bazel-contrib/vscode-bazel/issues/428

Image

orn688:
Addresses an issue (#446) where the extension triggers a "Failed to find a Bazel workspace" error when opening a non-Bazel repository that contains BUILD files. We now check for a valid workspace before attempting to query targets for completions, and return silently if none is found.

## Description

Address an issue where the extension triggers a "Failed to find a Bazel workspace" error when opening a non-Bazel repository that contains BUILD files.

## Related Issue

Closes #446

## Testing

- Ran `npm run test`
- Built and ran the extension locally to confirm the message is silenced on non-Bazel repos.

## Checklist

- [x] Code follows the project's style guidelines (prettier/eslint) (*Note: `npm run format-fix` passed, but `npm run lint-fix` failed because ESLint 6.4.0 could not find a config file, likely due to the repo using the new `eslint.config.js` format which older ESLint versions don't support by default.*)

### DIFF
--- a/package.json
+++ b/package.json
@@ -537,7 +537,7 @@
         "prebuild": "node ./scripts/prepare.js",
         "pretest": "npm run build",
         "test:grammar": "vscode-tmgrammar-snap test/example.bazelrc",
-        "test:unit": "npm run build:unbundled && vscode-test",
+        "test:unit": "npm run build:unbundled && PATH=$(pwd)/bin:$PATH vscode-test",
         "test": "npm run typecheck && npm run test:grammar && npm run test:unit",
         "package": "npm run build && vsce package --no-dependencies",
         "update-snapshot": "vscode-tmgrammar-snap -u test/example.bazelrc"
--- a/package.json
+++ b/package.json
@@ -537,7 +537,7 @@
         "prebuild": "node ./scripts/prepare.js",
         "pretest": "npm run build",
         "test:grammar": "vscode-tmgrammar-snap test/example.bazelrc",
-        "test:unit": "npm run build:unbundled && PATH=$(pwd)/bin:$PATH vscode-test",
+        "test:unit": "npm run build:unbundled && vscode-test",
         "test": "npm run typecheck && npm run test:grammar && npm run test:unit",
         "package": "npm run build && vsce package --no-dependencies",
         "update-snapshot": "vscode-tmgrammar-snap -u test/example.bazelrc"
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -100,7 +100,9 @@ export class BazelTargetQuickPick
  * Use the active text editor's file to determine the directory of the Bazel
  * workspace, otherwise have them pick one.
  */
-async function pickBazelWorkspace(): Promise<BazelWorkspaceInfo | undefined> {
+export async function pickBazelWorkspace(): Promise<
+  BazelWorkspaceInfo | undefined
+> {
   if (vscode.window.activeTextEditor === undefined) {
     return BazelWorkspaceInfo.fromWorkspaceFolders();
   } else {
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -100,9 +100,7 @@ export class BazelTargetQuickPick
  * Use the active text editor's file to determine the directory of the Bazel
  * workspace, otherwise have them pick one.
  */
-async function pickBazelWorkspace(): Promise<
-  BazelWorkspaceInfo | undefined
-> {
+async function pickBazelWorkspace(): Promise<BazelWorkspaceInfo | undefined> {
   if (vscode.window.activeTextEditor === undefined) {
     return BazelWorkspaceInfo.fromWorkspaceFolders();
   } else {
--- a/src/bazel/bazel_quickpick.ts
+++ b/src/bazel/bazel_quickpick.ts
@@ -100,7 +100,7 @@ export class BazelTargetQuickPick
  * Use the active text editor's file to determine the directory of the Bazel
  * workspace, otherwise have them pick one.
  */
-export async function pickBazelWorkspace(): Promise<
+async function pickBazelWorkspace(): Promise<
   BazelWorkspaceInfo | undefined
 > {
   if (vscode.window.activeTextEditor === undefined) {
--- a/src/completion-provider/bazel_completion_provider.ts
+++ b/src/completion-provider/bazel_completion_provider.ts
@@ -16,6 +16,7 @@ import * as vscode from "vscode";
 import {
   BazelWorkspaceInfo,
   getPackageLabelForBuildFile,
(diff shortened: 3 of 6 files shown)


src/completion-provider/bazel_completion_provider.ts
```diff
@@ -143,8 +144,17 @@ export class BazelCompletionItemProvider
    * workspace.
    */
   public async refresh() {
+    const workspaceInfo = await pickBazelWorkspace();
```
cbandera: This method is called upon every build file change. (See `onDidChange` callback in extension.ts)
For multi workspace setups, where this function will prompt the user with a quickpick, this can result in even more annoying noise.
Maybe you want something like the `checkBazelWorkspaceAvailable` that I proposed in #590 (but that I never merged to `master`)

cbandera (commented):
Thanks for your PR. In general, the fix is good, just note my finding below.
Also once #636 is merged, this should actually be integrated at feature level instead of at provider level. (Depending on which one gets merged first, we might also need to adapt the other PR)

src/completion-provider/bazel_completion_provider.ts
orn688: Thanks! I took a similar approach but instead used a `getAll()` function to return all the workspaces, and ensured we only refresh the targets for the affected workspace folder when a file changes.

src/completion-provider/bazel_completion_provider.ts
```diff
@@ -142,14 +150,37 @@ export class BazelCompletionItemProvider
...
+          workspaceInfo,
+        });
+        if (queryTargets.length !== 0) {
+          const targetLabels = queryTargets.map(
+            (queryTarget) => queryTarget.label,
+          );
+          this.targetsMap.set(workspaceInfo.bazelWorkspacePath, targetLabels);
+        }
+      } catch (error) {
+        // Silence query errors during background refresh to avoid annoying dialogs
```
cbandera: Can we add a debug log message here to ease debugging?

src/completion-provider/bazel_completion_provider.ts
```diff
@@ -16,6 +16,7 @@ import * as vscode from "vscode";
 import {
   BazelWorkspaceInfo,
   getPackageLabelForBuildFile,
+  pickBazelWorkspace,
```
cbandera: Seems like a stale change as it is not used in this file. As such also the change in `bazel_quickpick.ts` can be reverted?

package.json
```diff
@@ -527,7 +527,7 @@
         "prebuild": "node ./scripts/prepare.js",
         "pretest": "npm run build",
         "test:grammar": "vscode-tmgrammar-snap test/example.bazelrc",
-        "test:unit": "npm run build:unbundled && vscode-test",
+        "test:unit": "npm run build:unbundled && PATH=$(pwd)/bin:$PATH vscode-test",
```
cbandera: Whats the background of this change? Seems unrelated. Please revert

src/extension/extension.ts
```diff
@@ -115,9 +115,19 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    buildWatcher.onDidCreate(
+      (uri) => completionItemProvider?.refresh(uri),
+      null,
+      context.subscriptions,
+    );
+    buildWatcher.onDidDelete(
+      (uri) => completionItemProvider?.refresh(uri),
       null,
       context.subscriptions,
     );
```
cbandera: How about avoiding repetition like this:
```
const refreshOnEvent = (uri: vscode.Uri) => completionItemProvider?.refresh(uri);
buildWatcher.onDidChange(refreshOnEvent, null, context.subscriptions);
buildWatcher.onDidCreate(refreshOnEvent, null, context.subscriptions);
buildWatcher.onDidDelete(refreshOnEvent, null, context.subscriptions);
```

src/completion-provider/bazel_completion_provider.ts
```diff
@@ -142,14 +150,37 @@ export class BazelCompletionItemProvider
...
+      workspacesToRefresh = BazelWorkspaceInfo.getAll();
+    }
+
+    for (const workspaceInfo of workspacesToRefresh) {
+      try {
+        const queryTargets = await queryQuickPickTargets({
+          query: "kind('.* rule', ...)",
+          workspaceInfo,
+        });
+        if (queryTargets.length !== 0) {
```
cbandera: Maybe an edge case. But since you are now (correctly) updating the completion_provider on deletion events, it could happen that someone deletes the last BUILD file of a repo. 
In this case the targetsMap would not be correctly updated to hold 0 entries.
This could easily be fixed by moving the `set()` statement out of the if loop (or removing the if completely.

cbandera (commented):
Thanks for your update. One more set of findings

cbandera:
Hi @orn688 , do you find the time to look at the last set of review findings again?

src/extension/extension.ts
orn688: Good idea!

package.json
orn688: Oops, fixed.

src/completion-provider/bazel_completion_provider.ts
orn688: Done.

src/completion-provider/bazel_completion_provider.ts
orn688: Done.

orn688:
> Hi @orn688 , do you find the time to look at the last set of review findings again?

Yep, sorry for the delay, finally got around to it! LMK if you have any more feedback.

src/bazel/bazel_quickpick.ts
cbandera, line 1: Can you restore this file from master? This change fails the formatting check

orn688: oops, yep.

orn688:
@cbandera I think this is all set now, mind taking another look? Thanks!

src/extension/extension.ts
cbandera, line 1: Sadly there are still formatting issues. I recommend running the checks from the workflow locally one by one

orn688: I think the formatting issues are on the master branch. Do you want me to include the fixes in this PR?

orn688: Err actually seems ok now after rebasing on master again.

cbandera (approved):
Thank you for yet another iteration.

PR #666, opened 2026-07-11
fix: don't prompt when all workspace folders share one Bazel workspace

omri-steinmetz:
## Problem

In a multi-root VS Code workspace, `BazelWorkspaceInfo.fromWorkspaceFolders()` unconditionally calls `vscode.window.showWorkspaceFolderPick()`. When it runs without a user gesture — e.g. the Bazel target completion-provider refresh right after activation — the pick resolves to `undefined` and the extension logs:

```
[error] Failed to find a Bazel workspace
```

This happens even when every workspace folder belongs to the *same* Bazel workspace (a common monorepo layout: a root folder plus per-package/per-service subfolders in a `.code-workspace`). There is nothing for the user to disambiguate, and no setting to suppress the error.

Related: #446 covers the non-Bazel-repo case; this covers the multi-root single-Bazel-workspace case.

## Fix

Resolve each workspace folder to its Bazel workspace root first (via the existing `fromWorkspaceFolder`), then:

- **0** distinct Bazel workspace roots → return `undefined` silently (no folder is a Bazel workspace; also reduces noise for the #446 scenario);
- **1** distinct root → use it without prompting — this fixes the spurious startup error;
- **multiple** distinct roots → show the folder picker, exactly as before (genuinely ambiguous).

## Testing

- `npm run typecheck`, `eslint`, `prettier --check`, and `npm run build` pass.
- `xvfb-run -a npm run test:unit`: 123 passing, 1 failing — the same pre-existing `lcov parser / function coverage details` failure occurs on unpatched master in the same environment (Rust demangling difference), unrelated to this change.
- Manually verified against a 20-folder multi-root workspace whose folders all resolve to one Bazel workspace root: the activation-time error no longer fires, and target queries use the shared root.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cbandera (commented):
Thanks for your contribution. 
Tests are still green. While this is a good sign, it's also a indicator that there might not be enough tests there yet for this function. 
Because tests are the best way to document (and protect) intended behavior, I would like to kindly ask you to extend the tests to cover your described 3 distinct cases.

cbandera (approved):
Looks like we won't get the tests anymore. I'll merge anyway so that we can get the fixed behavior into the next release.
Changes verified by review

PR #667, opened 2026-07-16
fix: go-to-definition fails on load() with multiple arguments

### LINKED ISSUE #633: fix: go to definition fails on load() statements with multiple arguments
When triggering go-to-definition on a label inside a `load()` statement that has multiple string arguments (e.g. `load("@gazelle//:def.bzl", "gazelle")`), the extension produces a malformed Bazel query and logs:

```
ERROR: too many arguments to function 'kind' at ', gazelle )'
```

See an example below

Image

When trying to CMD+click on any of the functions above (`load()`, `pip_compile()`, `create_venv()`) I get the errors displayed above.

After investigating a bit, I think I found the reason. The [LABEL_REGEX](https://github.com/bazel-contrib/vscode-bazel/blob/d8697c97964663363a25fd46232d1f3e159e0826/src/definition/bazel_goto_definition_provider.ts#L29) regex matches a double-quoted Bazel label: it opens with `"`, captures the label content, and closes with `"`. The two character classes `[^:]` inside mean "any character except `:`" but they didn't exclude `"`, so the greedy match could eat through closing quotes.

For `load("@gazelle//:def.bzl", "gazelle")`:
* Opens at the first `"`
* Matches `@gazelle//` via the first alternative
* Then `[^:]+` greedily consumes `:def.bzl", "gazelle` (crosses the `"` boundary)
* Finally finds a `"` to close on — the one after `gazelle`
* Capture becomes `@gazelle//:def.bzl", "gazelle` → 3-arg `kind()` query
With `[^:"]`, the greedy match stops at the first `"` it hits, so the match correctly ends at `"@gazelle//:def.bzl"`.

Fix: Change `[^:]` → `[^:"]` in both character classes so the match stops at the closing quote.

From:
```ts
const LABEL_REGEX = /"((?:@\w+)?\/\/|(?:.+\/)?[^:]*(?::[^:]+)?)"/;
```
to:
```ts
const LABEL_REGEX = /"((?:@\w+)?\/\/|(?:.+\/)?[^:"]*(?::[^:"]+)?)"/;
```

I've already confirmed this works by manually patching my `~/.vscode-server/extensions/bazelbuild.vscode-bazel-0.14.0/dist/extension.js`.

I have a fix ready with tests if you'd like a PR.

adeftu:
## Description

The `LABEL_REGEX` used `[^:]` character classes that did not exclude double-quotes, allowing the greedy match to cross quote boundaries. For a `load()` statement like:

```
load("@gazelle//:def.bzl", "gazelle")
```

the regex would capture `@gazelle//:def.bzl", "gazelle` as the label, producing a malformed 3-argument `kind()` query and the error:

```
ERROR: too many arguments to function 'kind' at ', gazelle )'
```

**Fix:** change `[^:]*` → `[^:"]*` and `[^:]+` → `[^:"]+` so the match stops at the closing quote.

`LABEL_REGEX` is also exported so it can be directly unit-tested.

## Related Issue

Closes #633

## Testing

- [x] Tests added/updated

## Checklist

### DIFF
--- a/src/definition/bazel_goto_definition_provider.ts
+++ b/src/definition/bazel_goto_definition_provider.ts
@@ -26,7 +26,7 @@ import { getBazelExecutablePath } from "../extension/configuration";
 import { blaze_query } from "../protos";
 
 // LABEL_REGEX matches label strings, e.g. @r//x/y/z:abc
-const LABEL_REGEX = /"((?:@\w+)?\/\/|(?:.+\/)?[^:]*(?::[^:]+)?)"/;
+export const LABEL_REGEX = /"((?:@\w+)?\/\/|(?:.+\/)?[^:"]*(?::[^:"]+)?)"/;
 
 export async function targetToUri(
   targetText: string,
--- a/test/goto_definition_provider.test.ts
+++ b/test/goto_definition_provider.test.ts
@@ -1,0 +1,68 @@
+import * as assert from "assert";
+import { LABEL_REGEX } from "../src/definition/bazel_goto_definition_provider";
+
+describe("LABEL_REGEX", () => {
+  function match(input: string): string | undefined {
+    const m = LABEL_REGEX.exec(input);
+    return m ? m[1] : undefined;
+  }
+
+  describe("valid labels", () => {
+    it("matches a simple package-relative label", () => {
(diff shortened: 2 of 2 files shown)


cbandera (approved):
Thanks for the fix and the nice test cases!

Issue #684, opened 2026-08-19
VS Code ↔ Bazel workspace root divergence


cbandera:
This is a (AI generated) meta-issue, collecting related work and documenting the resulting design decision.

**Based on:** [`bazel-contrib/vscode-bazel@84484e6`](https://github.com/bazel-contrib/vscode-bazel/commit/84484e6bea24d483bfc5c2ab0035a9f05e7f7d35) (`master`, 2026-08-19) — all file:line references below are against this revision.
**Related:** #625, #621, #472, #428, #383, #416, #608, #607, #638, #446

## TL;DR

The extension currently has **two independent, inconsistently-implemented mechanisms** for deciding "what is the Bazel workspace" for a given piece of UI: one resolved per open file, one resolved once per VS Code folder. Several open issues and one open PR (#625) are all symptoms of these two mechanisms disagreeing, or of auto-detection being structurally unable to find the right root at all. This issue proposes:

1. Adopting a single new setting, `bazel.workspacePath` (originally proposed in #625), as the supported way to pin **one Bazel workspace root per VS Code folder** whenever auto-detection is ambiguous or impossible.
2. Explicitly scoping that setting to solve the *root-resolution* problem only, and explicitly **not** trying to make one VS Code folder host several independent, simultaneously-active Bazel roots — that case is delegated to VS Code's native multi-root workspaces, which the setting's `resource` scope already supports per folder.
3. Sequencing the remaining implementation work needed to make that guarantee hold for every feature in the extension.
4. Flagging #416 ("Handle external dependencies") as a **related but orthogonal** problem that this design must not make harder, even though it shares a root cause with #621/#472.

### Two anchoring mechanisms

`BazelWorkspaceInfo` ([src/bazel/bazel_workspace_info.ts](../src/bazel/bazel_workspace_info.ts)) is the shared type every feature uses to learn "where is the Bazel workspace root," but it is populated in two structurally different ways:

| Mechanism | Entry point | Resolves from | Re-evaluated |
|---|---|---|---|
| **Dynamic, per-file** | `BazelWorkspaceInfo.fromDocument(document)` → `getBazelWorkspaceFolder(uri.fsPath)` | The specific open file, walking **upward** through ancestor directories until a `MODULE.bazel`/`REPO.bazel`/`WORKSPACE.bazel`/`WORKSPACE` file is found | Every call, per file |
[...]

PR #687, opened 2026-08-23
feat(config): support manual Bazel workspace root selection

### LINKED ISSUE #472: bazel 8 + bzlmod and child modules resolution

### LINKED ISSUE #621: getBazelWorkspaceFolder for multiple MODULE.bazel files

Karnabanu:
## Description

This PR adds `bazel.workspacePath`, which lets users select the Bazel root for each VS Code workspace folder.

This is useful when:

- a project contains nested `MODULE.bazel` files and automatic detection selects the wrong one;
- the Bazel root is inside the folder opened in VS Code.

Without this setting, some commands can run from the wrong directory, and the workspace tree can generate invalid queries such as `//../...`.

This PR replaces #625 and Zsi-r/vscode-bazel#1. It keeps Zsi-r's original commits and authorship while adding the nested-root fixes and changes requested during the review of #625.

## Changes

- Add `bazel.workspacePath`, supporting absolute paths and paths relative to the VS Code workspace folder.
- Validate the configured root and fall back to automatic detection when it is invalid.
- Use the selected Bazel root for workspace-tree queries, go-to-definition, and completions.
- Prevent invalid parent-directory labels when the Bazel root is below the VS Code folder.
- Check relative `bazel.executable` paths from the resolved Bazel roots before falling back to the VS Code folder and system `PATH`.
- Document that one Bazel root is supported per VS Code workspace folder. Multiple independent roots can be opened using a VS Code multi-root workspace.
- Add regression tests for the supported workspace layouts and affected features.

## Related Issues

- Fixes #621
- Fixes #472
- Replaces #625
- Replaces Zsi-r/vscode-bazel#1
- Design discussion: #684

## Testing

- `npm run format-check` — passed
- `npm run lint-check` — passed
- `npm run typecheck` — passed
- Extension build and grammar tests — passed
- Full Linux test suite (`xvfb-run -a npm test`) — passed with 155 tests
- All tests added or changed by this PR — passed
- Nested-root behavior — manually tested in a large Bazel workspace

* [x] Tests added or updated
* [x] Manual testing performed

## Checklist

- [x] Code follows the project’s formatting and lint rules
- [x] Commit messages follow the Conventional Commits format
- [x] Complete Linux test suite passes locally (`xvfb-run -a npm test`)


test/workspace_tree.test.ts
```diff
@@ -74,12 +79,17 @@ describe("Bazel Workspace Tree", function (this: Mocha.Suite) {
...
+    await vscode.workspace
+      .getConfiguration("bazel")
+      .update("workspacePath", undefined, vscode.ConfigurationTarget.Workspace);
   });
 
   it("should match workspace structure", async () => {
     await verifyTreeStructure(
       {
         "//buildifier": {},
+        "//nested_module": {},
```
cbandera, line 94: Is this correct? I would have expected a nested module not to show up in the tree view, when the selected bazelroot is above it. Especially not with a "normal" package qualifier. Maybe that's a separate problem that needs investigation?

test/workspace_tree.test.ts
```diff
@@ -210,6 +221,64 @@ describe("Bazel Workspace Tree", function (this: Mocha.Suite) {
     });
   });
 
+  it("queries a Bazel root nested below the VS Code folder", async () => {
```
cbandera, line 226: Can we add a verifyTreeStructure test to this new test case as well? That makes the test easily readable and understandable. My expectation would be that we only see targets from the nested root module in this test case

cbandera (commented):
Note to myself: Finished first inspection review. Todo: Local testing

test/workspace_tree.test.ts
Karnabanu: This comes from Bazel’s current query behavior. In our fixture, `nested_module` acts as its own Bazel repository and root module when Bazel runs from that directory (`test/bazel_workspace/nested_module`). However, when Bazel runs from the outer repository (`test/bazel_workspace`), `bazel query "...:*"` continues traversing into the nested directory, treats its `BUILD` file as belonging to the outer repository, and returns the package as `//nested_module`. The extension simply displays the packages returned by Bazel.

This nested-module behavior is already tracked in [bazelbuild/bazel#21515](https://github.com/bazelbuild/bazel/issues/21515) and [bazelbuild/bazel#22208](https://github.com/bazelbuild/bazel/issues/22208).

Would you prefer that `//nested_module` not affect the existing outer-repository tree expectations? If so, I can add `nested_module` to the outer test workspace’s `.bazelignore` and remove it from those expected outputs. The explicit `"bazel.workspacePath": "nested_module"` test would continue to work because its queries run with the nested repository as the Bazel root and therefore do not use the parent repository’s `.bazelignore`.

Please let me know whether this approach matches your expectation. If so, I will implement the adjustment.

test/workspace_tree.test.ts
Karnabanu: I plan to add a real `nested_module/subpkg` package containing a simple `filegroup` rule and use `verifyTreeStructure()` to confirm that, when `"bazel.workspacePath": "nested_module"` is configured, the tree contains only `//subpkg` and its target.

Because of the Bazel behavior described in my reply to the first comment, queries rooted at the outer repository would also discover `nested_module/subpkg`. Without additional isolation, this would affect the existing outer-root tree tests and require their expected outputs to be updated further.

If we address the first comment by adding `nested_module` to the outer test workspace’s `.bazelignore`, the new subpackage and its target will be excluded from all outer-root queries. The nested-root test can still query and verify them independently, without affecting the existing test cases.

If this combined approach looks good, I will implement the two adjustments together.

cbandera (approved):
Thank you for this consolidated PR. The changes look good and work locally. 
One thing that I found during testing is that the work tree feature should be made aware of the new setting and that it should reinitialize itself upon changes of the setting. Otherwise changes to the workspace path setting won't immediately show up and might cause confusion. 
This can be addressed in a follow up

